PR #3, opened 2018-11-21
Handle strings returned from Z3 as borrowed.

badboy:
z3.rs doesn't actually own the memory, but wrapping it in a CString will
assume it does and drop the content (and thus free it) when going out of scope.

The new test will crash with a SIGSEGV on the old code.


badboy:
I'm not 100% sure, but this might actually now leak the strings, as Z3 allocated them.

waywardmonkeys:
Thanks!

PR #6, opened 2018-12-31
Don't move the output of Z3_solver_to_string()

sdamashek:
We don't own the string returned by Z3_solver_to_string(). Using CString causes a move and subsequent double free, resulting in a segfault. This commit changed fmt() to use a CStr like in Ast.


waywardmonkeys:
Thanks!

waywardmonkeys:
I published a v0.3.1 with this change.

PR #12, opened 2019-02-02
z3: tests: Fix test_pb_ops_model for Z3 >= 4.8.3

Armavica:
Z3 4.8.1 returned `x=true, y=false`, whereas Z3 4.8.3 and 4.8.4 return
`x=false, y=true`.  Both models are valid but the assertion `xv > yv`
selects only for the former.


waywardmonkeys:
Thanks!

PR #14, opened 2019-03-08
Add API support for AST translate.

radustoenescu:
High-level API support for `Z3_translate` operation.

Usage example can be found in the corresponding unit test.

### DIFF
--- a/z3/tests/lib.rs
+++ b/z3/tests/lib.rs
@@ -119,6 +119,26 @@ fn test_ast_translate() {
 }
 
 #[test]
+fn test_solver_translate() {
+    let cfg = Config::new();
+    let source = Context::new(&cfg);
+    let a = source.named_int_const("a");
+
+    let destination  = Context::new(&cfg);
+    let translated_a = a.translate(&destination);
+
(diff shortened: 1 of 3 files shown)


z3/tests/lib.rs
```diff
@@ -101,6 +101,43 @@ fn test_format() {
...
+//    let slv = Solver::new(&destination);
+//    slv.assert(&translated_a._eq(&destination.from_u64(2)));
+//    assert!(slv.check());
+//
+//    let translated_slv = slv.translate(&source);
+//
+//    slv.assert(&translated_a._eq(&destination.from_u64(3)));
+//    assert!(! slv.check());
+//    assert!(translated_slv.check());
+//}
```
waywardmonkeys: Why not enable this test or combine it with the one above?

radustoenescu:
I've uncommented the test (had some z3-binary problems on my machine that caused the test to fail).

In the meantime, I also added the solver translate method.

Let me know what you think.

waywardmonkeys:
Sorry for the delay in dealing with this. Looks good to me and thanks for the contribution!

PR #15, opened 2019-04-25
Use bindgen for generating bindings

jrakow:


waywardmonkeys:
I've been thinking about this for the past day.

What's your end goal here?

I've long been tempted to instead have `z3-sys` build a copy of Z3 ... but that's also annoying.

I wish that Z3 itself was more semver focused, but it isn't ...

jrakow:
I tried using the z3 bindings with the newest version of z3, but that failed, because 4.8 is incompatible with 4.7. I would like it so that z3 either works with all recent versions (infeasible) or just installs a working version. I also did not find information on what versions the bindings are compatible with.

I have now added z3-4.7.1 as a submodule in the `vendor` directory. The build script has to copy it to the output directory, because z3's scripts add files in the source tree. It is then built there and the bindings are generated by bindgen. When the submodule is changed, only necessary files are copied and rebuilt.

The bindings always work with the installed version. This project can just update the submodule for future versions of z3.

PR #17, opened 2019-05-27
Add bindings for FuncDecl

jrakow:

PR #18, opened 2019-05-27
fix FuncDecl

jrakow:
Followup to #17


waywardmonkeys:
I need to go to the gym this morning (now), but when I'm back, I'll go through all of these! Thanks!

PR #30, opened 2019-06-20
Create bitvectors from constants

cdisselkoen:
Unless I'm missing something, the high-level bindings have no simple way to create bitvector constants.  This PR adds new functions `Ast::bitvector_from_i64()` and `Ast::bitvector_from_u64()` to mirror the existing `Ast::from_i64` and `Ast::from_u64()` which produce integer constants.  Also adds a test using the new functions (actually, no existing tests used bitvectors at all).

Happy to take feedback on anything - function names, function arguments, anything else.  I'm fairly new here.

All the tests in `tests/lib.rs` still pass with this patch.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -123,6 +123,22 @@ impl<'ctx> Ast<'ctx> {
         })
     }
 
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
Thanks! One day, I'd like to go through the Python docs and write Rust versions of the various examples ...

Issue #2, opened 2018-11-21
fresh_const (and friends) are broken


badboy:
Currently the [semver test](z3/tests/semver_tests.rs) fails due to a version mismatch in the result.

I rewrote the test using `ctx.numbered_int_const` instead of `ctx.fresh_int_const` and it now passes (changes in https://gist.github.com/badboy/e126802b474d7de60d5d73e47531b01c/revisions)

I had a similar problem in my own code and also used `numbered_int_const` there.

badboy:
Simplified failing test:

```rust
#[test]
fn test_fresh_int_const() {
    let _ = env_logger::try_init();
    let cfg = Config::new();
    let ctx = Context::new(&cfg);
    let x = ctx.fresh_int_const("x");
    let zero = ctx.from_i64(0);
    let solver = Solver::new(&ctx);
    solver.assert(&x.gt(&zero));
    assert!(solver.check());

    let model = solver.get_model();
    let xv = model.eval(&x).unwrap().as_i64().unwrap();
    assert_eq!(xv, 1);
}
```

Output:

```
thread 'test_fresh_int_const' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`', z3/tests/lib.rs:106:5
```

badboy:
[This C code example](https://gist.github.com/badboy/af4870a56078a50373edebc854c7f4c5) has the same problem.

badboy:
As can be seen [in this example](https://gist.github.com/badboy/1cc6d05ce9804c41ef9110daa62f45ee) the fresh const is not even part of the model, so I guess that's not how `fresh_const` should be used anymore?

waywardmonkeys:
I'll try to look at this some today as well! Thanks!

waywardmonkeys:
It is interesting to look at `Z3_solver_to_string` output from your C example:

```
(declare-fun k!1 () Bool)
(declare-fun x!0 () Int)
(declare-fun k!2 () Bool)
(declare-fun y () Int)
(declare-fun k!3 () Bool)
(declare-fun k!2 () Int)
(declare-fun k!4 () Bool)
(declare-fun k!5 () Bool)
(declare-fun k!6 () Bool)
(assert (> x!0 0))
(assert (> y 0))
(assert (> k!2 0))
(model-del k!1)
(model-add x!0 () Int (ite k!1 0 (+ 0 1)))
(model-del k!2)
[...]

PR #32, opened 2019-06-27
AST subtypes

### LINKED ISSUE #31: Question/discussion/suggestion: AST subtypes
Trying to combine `Ast`s of different `Sort`s often results in errors.  For instance, you can't `add` an integer and a bitvector, you can only `select` on an array, etc.  In fact, the `Ast` type appears to have a number of distinct groups of methods, only one of which applies to any given `Ast` (depending on its `Sort`): the first group (`add`, `sub`, etc) which apply to integers (and maybe reals? Not sure as I never use reals); the second group (`bv*`) which apply to bitvectors; `select` and `store` which apply to arrays; `set_*` which apply to sets; and `pb_*` (I'm not sure what these are either).

Since each of these groups of methods apply to different kinds of `Ast`s, and rarely can different kinds of `Ast`s be mixed without errors, I was wondering if any thought has been given to splitting `Ast` into a number of subtypes, perhaps `IntAst`, `BVAst`, `ArrayAst`, etc (obviously different names could be used).  This would not only clarify exactly which `Ast` operations apply to each type, it would also help prevent more errors statically at the Rust type level, thus making a safer set of high-level bindings.

I'm potentially willing to help implement this if it's desired and feasible, but I wanted to ask first if it's a reasonable suggestion, and get feedback from anyone with more experience with Z3 (and Rust) than me.  Is there a reason this is a bad idea?

cdisselkoen:
This is a first effort to try to address #31.  Comments or code review very welcome.

There are a number of questions or ambiguities, which I've called out in code comments for now (and can remove once the questions are resolved).  For instance, with AST subtypes, I think it makes sense to remove (or at least deprecate) the generic `ctx.named_const()`, `ctx.fresh_const()`, etc methods in favor of the typed variants `ctx.named_bool_const()`, `ctx.named_int_const`, etc; but I'm not sure how important the generic variants are or how breaking that would be.

The first commit in this PR exposes a few additional functions in the high-level API that were previously only in `z3-sys`.  In particular, this means the `SortKind` enum is now a public part of the `z3` crate.  I'm assuming that's OK, but I wanted to call it out specifically just in case.  Also, if that's enough of an unrelated change to be a separate PR, I'm happy to pull it out separately.

All current tests are passing (except `test_solve_simple_semver_example` which was already failing on `master` (#2?)), but arguably many more tests would be needed to ensure that all these method types are actually correct.  I wanted to get some feedback first though before writing a bunch more tests.

For crate users, these are clearly breaking changes for the API, and would need to be semver'd appropriately if this PR is accepted.

If this PR is going to be accepted, I'll squash before it actually gets merged, in order to keep the git history clean.

Fixes #31.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -1,52 +1,127 @@
 use std::cmp::{Eq, PartialEq};
-use std::convert::TryInto;
+use std::convert::{TryInto, TryFrom};
 use std::ffi::{CStr, CString};
 use std::fmt;
 use std::hash::{Hash, Hasher};
 use z3_sys::*;
-use Ast;
 use Context;
 use Sort;
 use Symbol;
 use Z3_MUTEX;
 
+/// [`Ast`](trait.Ast.html) node representing a boolean value.
+pub struct Bool<'ctx> {
+    pub(crate) ctx: &'ctx Context,
+    pub(crate) z3_ast: Z3_ast,
+}
+
+/// [`Ast`](trait.Ast.html) node representing an integer value.
+pub struct Int<'ctx> {
+    pub(crate) ctx: &'ctx Context,
+    pub(crate) z3_ast: Z3_ast,
+}
+
+/// [`Ast`](trait.Ast.html) node representing a real value.
+pub struct Real<'ctx> {
+    pub(crate) ctx: &'ctx Context,
+    pub(crate) z3_ast: Z3_ast,
+}
+
+/// [`Ast`](trait.Ast.html) node representing a bitvector value.
+pub struct BV<'ctx> {
+    pub(crate) ctx: &'ctx Context,
+    pub(crate) z3_ast: Z3_ast,
+}
+
+/// [`Ast`](trait.Ast.html) node representing an array value.
+/// An array in Z3 is a mapping from indices to values.
+pub struct Array<'ctx> {
+    pub(crate) ctx: &'ctx Context,
+    pub(crate) z3_ast: Z3_ast,
+}
+
+/// [`Ast`](trait.Ast.html) node representing a set value.
+pub struct Set<'ctx> {
+    pub(crate) ctx: &'ctx Context,
+    pub(crate) z3_ast: Z3_ast,
+}
+
+// These `From` impls imply the corresponding `Into` impls
+impl<'ctx> From<Bool<'ctx>> for Z3_ast {
+    fn from(ast: Bool<'ctx>) -> Self {
+        ast.z3_ast
+    }
+}
+
+impl<'ctx> From<Int<'ctx>> for Z3_ast {
+    fn from(ast: Int<'ctx>) -> Self {
+        ast.z3_ast
+    }
+}
+
+impl<'ctx> From<Real<'ctx>> for Z3_ast {
+    fn from(ast: Real<'ctx>) -> Self {
+        ast.z3_ast
+    }
+}
+
+impl<'ctx> From<BV<'ctx>> for Z3_ast {
+    fn from(ast: BV<'ctx>) -> Self {
+        ast.z3_ast
+    }
+}
+
+impl<'ctx> From<Array<'ctx>> for Z3_ast {
+    fn from(ast: Array<'ctx>) -> Self {
+        ast.z3_ast
+    }
+}
+
+impl<'ctx> From<Set<'ctx>> for Z3_ast {
+    fn from(ast: Set<'ctx>) -> Self {
+        ast.z3_ast
+    }
+}
+
 macro_rules! unop {
-    ( $f:ident, $z3fn:ident ) => {
-        pub fn $f(&self) -> Ast<'ctx> {
-            Ast::new(self.ctx, unsafe {
+    ( $f:ident, $z3fn:ident, $retty:ty ) => {
+        pub fn $f(&self) -> $retty {
+            <$retty>::new(self.ctx, unsafe {
                 let guard = Z3_MUTEX.lock().unwrap();
                 $z3fn(self.ctx.z3_ctx, self.z3_ast)
             })
-    }
+        }
     };
 }
 
 macro_rules! binop {
-    ( $f:ident, $z3fn:ident ) => {
-        pub fn $f(&self, other: &Ast<'ctx>) -> Ast<'ctx> {
-            Ast::new(self.ctx, unsafe {
+    ( $f:ident, $z3fn:ident, $retty:ty ) => {
+        pub fn $f(&self, other: &Self) -> $retty {
+            <$retty>::new(self.ctx, unsafe {
                 let guard = Z3_MUTEX.lock().unwrap();
                 $z3fn(self.ctx.z3_ctx, self.z3_ast, other.z3_ast)
             })
-    }
+        }
     };
 }
 
+/* We aren't currently using the trinop! macro for any of our trinops
 macro_rules! trinop {
-    ( $f:ident, $z3fn:ident ) => {
-        pub fn $f(&self, a: &Ast<'ctx>, b: &Ast<'ctx>) -> Ast<'ctx> {
-            Ast::new(self.ctx, unsafe {
+    ( $f:ident, $z3fn:ident, $retty:ty ) => {
+        pub fn $f(&self, a: &Self, b: &Self) -> $retty {
+            <$retty>::new(self.ctx, unsafe {
                 let guard = Z3_MUTEX.lock().unwrap();
                 $z3fn(self.ctx.z3_ctx, self.z3_ast, a.z3_ast, b.z3_ast)
             })
-    }
+        }
     };
 }
+*/
 
 macro_rules! varop {
-    ( $f:ident, $z3fn:ident ) => {
-        pub fn $f(&self, other: &[&Ast<'ctx>]) -> Ast<'ctx> {
-            Ast::new(self.ctx, unsafe {
+    ( $f:ident, $z3fn:ident, $retty:ty ) => {
+        pub fn $f(&self, other: &[&Self]) -> $retty {
+            <$retty>::new(self.ctx, unsafe {
                 let guard = Z3_MUTEX.lock().unwrap();
                 let mut tmp = vec![self.z3_ast];
                 for a in other {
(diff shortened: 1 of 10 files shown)


waywardmonkeys:
Does the `SortKind` need to be exposed outside of `z3` or just used by `z3`?

z3/src/lib.rs
```diff
@@ -55,19 +56,12 @@ pub struct Symbol<'ctx> {
     z3_sym: Z3_symbol,
 }
 
-/// Kind of [`Ast`](struct.Ast.html) used to represent types.
+/// Sorts represent the various 'types' of [`Ast`s](struct.Ast.html).
```
waywardmonkeys: This `struct.Ast.html` probably should become `trait.Ast.html`, right?

waywardmonkeys:
Having an `Into` / `From` impl to convert from the various AST types to the raw `Z3_ast` might be prettier than some of the current code. Just a thought ...

z3/src/lib.rs
waywardmonkeys: You got that in a followup commit already.

waywardmonkeys:
This is a good start ... there are some other nits that I have, like perhaps we can have `z3::ast::Bool` instead of `z3::BoolAst` ... and clean up the naming that way. But would you mind force pushing to this PR a cleaned up set of commits and perhaps we'll just land it and we can iterate on it from there?

I've had a terrible headache for the last couple of days, so I'm going to sign off for now and get some rest.

cdisselkoen:
> Does the `SortKind` need to be exposed outside of `z3` or just used by `z3`?

I had added a method `get_kind()` on `Sort` which returns `SortKind`, but I can remove it.  `z3` can just use `z3_sys`'s `Z3_get_sort_kind` when it needs the `SortKind`.  This way `SortKind` can stay internal.

cdisselkoen:
Changelog for that force-push:

- Removed `Sort::get_kind()` and no longer expose `z3_sys::SortKind`
- Renamed `BoolAst` to `ast::Bool`, etc. `ast` is now a `pub mod` on `z3`.
- Added `From` impls to convert the various AST types to `Z3_ast`; these imply the corresponding `Into` impls
- Squashed into a single commit

cdisselkoen:
Agh, I apologize.  I'm new enough to Rust that I didn't realize that `cargo test` wouldn't necessarily run all tests before quitting; I need `cargo test --no-fail-fast`.  There are a bunch more failing tests, which I will fix before this is merged.

cdisselkoen:
Force-pushed again.  Added some more `From` impls.  Only two doctests failing now; one I think we should just accept as a regression for the moment (the `enumeration_sort` test fails with `not yet implemented: Dynamic with kind Datatype`), and the other has me stumped (the `ast::BV::from_int()` test fails due to `as_i64()`  returning `None` on a `BV`).

cdisselkoen:
Fixed `ite`, which should be `(Bool, T, T) -> T` rather than `(Bool, T, T) -> Bool`

z3/src/solver.rs
```diff
@@ -205,9 +203,12 @@ impl<'ctx> Solver<'ctx> {
     /// - [`Config::set_proof_generation()`](struct.Config.html#method.set_proof_generation)
     ///
     /// [proof generation is not enabled]: struct.Config.html#method.set_proof_generation
-    pub fn get_proof(&self) -> Ast<'ctx> {
+    //
+    // TODO: what kind of AST does this actually return? If it's consistent, we can just
+    // make that the return type.
```
waywardmonkeys: In the Python bindings, this is labelled as an `ExprRef`, where the inheritance hierarchy is `AstRef` -> `ExprRef` -> (`BoolRef`, etc...).

But could we use an `-> impl Ast<'ctx>` here instead of `ast::Dynamic`?

cdisselkoen: `-> impl Ast<'ctx>` seems to work in my testing.  In my simple example, the returned `Ast` had the sort `SortKind::Unknown`.

cdisselkoen: Of course, we still have to use some concrete constructor in the method body.  In my testing, I was still calling `ast::Dynamic::new()` to construct the return value, so it's not terribly surprising I guess that the `-> impl Ast` worked as a declaration for the return type.

waywardmonkeys: In Python, they use `_to_expr_ref`:

```
def _to_expr_ref(a, ctx):
    if isinstance(a, Pattern):
        return PatternRef(a, ctx)
    ctx_ref = ctx.ref()
    k = Z3_get_ast_kind(ctx_ref, a)
    if k == Z3_QUANTIFIER_AST:
        return QuantifierRef(a, ctx)
    sk = Z3_get_sort_kind(ctx_ref, Z3_get_sort(ctx_ref, a))
    if sk == Z3_BOOL_SORT:
        return BoolRef(a, ctx)
    if sk == Z3_INT_SORT:
        if k == Z3_NUMERAL_AST:
            return IntNumRef(a, ctx)
        return ArithRef(a, ctx)
    if sk == Z3_REAL_SORT:
        if k == Z3_NUMERAL_AST:
            return RatNumRef(a, ctx)
        if _is_algebraic(ctx, a):
            return AlgebraicNumRef(a, ctx)
        return ArithRef(a, ctx)
    if sk == Z3_BV_SORT:
        if k == Z3_NUMERAL_AST:
            return BitVecNumRef(a, ctx)
        else:
            return BitVecRef(a, ctx)
    if sk == Z3_ARRAY_SORT:
        return ArrayRef(a, ctx)
    if sk == Z3_DATATYPE_SORT:
        return DatatypeRef(a, ctx)
    if sk == Z3_FLOATING_POINT_SORT:
        if k == Z3_APP_AST and _is_numeral(ctx, a):
            return FPNumRef(a, ctx)
        else:
            return FPRef(a, ctx)
    if sk == Z3_FINITE_DOMAIN_SORT:
        if k == Z3_NUMERAL_AST:
            return FiniteDomainNumRef(a, ctx)
        else:
            return FiniteDomainRef(a, ctx)
    if sk == Z3_ROUNDING_MODE_SORT:
        return FPRMRef(a, ctx)
    if sk == Z3_SEQ_SORT:
        return SeqRef(a, ctx)
    if sk == Z3_RE_SORT:
        return ReRef(a, ctx)
    return ExprRef(a, ctx)
```

But for the proof to be generated, the `check` / `check_assumptions` must have failed and proof generation must be enabled on the context.  In C, you'd do that via:

```
    Z3_set_param_value(cfg, "proof", "true");
```

waywardmonkeys:
Also, thanks for doing this! The work is definitely appreciated and a great step forwards.

I suspect that we're almost at a place where we can merge it and then work on improving it from there...

z3/src/solver.rs
cdisselkoen: Right, which is called by `Config::set_proof_generation()` in Rust if I'm not mistaken.

So I guess this really is an `Ast` of no specialized type then.  We could create a minimal `Ast` subtype for this, we could even call it `Proof` if you want.  Is that the best way forwards?

waywardmonkeys: Let's leave it as is for now, I guess and then we can iterate on it and experiment and think.

waywardmonkeys:
Apart from `Dynamic` ... is there anything blocking merging this?

cdisselkoen:
What specifically about `Dynamic` do you want addressed?

I'll make a few more comments inline about some other small issues.

z3/src/ast.rs
```diff
@@ -381,75 +381,975 @@ impl<'ctx> Ast<'ctx> {
...
+    binop!(modulo, Z3_mk_mod, Self);
+    binop!(power, Z3_mk_power, Self);
+    unop!(unary_minus, Z3_mk_unary_minus, Self);
+    binop!(lt, Z3_mk_lt, Bool<'ctx>);
+    binop!(le, Z3_mk_le, Bool<'ctx>);
+    binop!(gt, Z3_mk_gt, Bool<'ctx>);
+    binop!(ge, Z3_mk_ge, Bool<'ctx>);
+    // Z3 does support mixing ints and reals in add(), sub(), mul(), div(), and power()
+    //   (but not rem(), modulo(), lt(), le(), gt(), or ge()).
+    // TODO: we could consider expressing this by having a Numeric trait with these methods.
```
cdisselkoen: There's this question about mixing ints and reals in certain methods (which Z3 supports but the current draft of this PR does not)

cdisselkoen: (see the rest of this comment as well, continues to line 527)

z3/src/context.rs
```diff
@@ -94,6 +94,9 @@ impl Context {
         Symbol::from_string(self, s)
     }
 
+/* Deprecate in favor of the typed variants below?
+    Or even better, the {Bool,Int,etc}Ast::new_const() and ::fresh_const() methods?
+
```
cdisselkoen: this question about deprecating the untyped methods on `Context`

z3/src/context.rs
```diff
@@ -105,69 +108,71 @@ impl Context {
...
+    pub fn from_bool(&self, b: bool) -> ast::Bool {
+        ast::Bool::from_bool(self, b)
     }
 
-    pub fn from_u64(&self, u: u64) -> Ast {
-        Ast::from_u64(self, u)
+    // Deprecate from_u64() and from_i64() in favor of
+    // ast::Int::from_u64()/from_i64() and ast::BV::from_u64()/from_i64()
+    pub fn from_u64(&self, u: u64) -> ast::Int {
+        ast::Int::from_u64(self, u)
```
cdisselkoen: this is another deprecation question

z3/src/sort.rs
```diff
@@ -212,6 +212,8 @@ impl<'ctx> Sort<'ctx> {
             Z3_mk_int64(self.ctx.z3_ctx, i, self.z3_sort)
         })
     }
+*/
+
```
cdisselkoen: I commented out the `Sort::from_u64(&self, u: u64)` and `Sort::from_i64(&self, i: i64)` methods; if you're okay with that, we can just remove them before merging rather than leaving them commented out

cdisselkoen:
and then my final comment is that we should probably add more doctests for most of the methods.  For instance, I accidentally had the wrong type for `ite` at first until I fixed it earlier today, as noted above in the thread.  But we can probably merge now and add tests later if you're comfortable with that.

z3/src/sort.rs
waywardmonkeys: Okay to remove, I think.

z3/src/context.rs
waywardmonkeys: Deprecate / remove.

z3/src/context.rs
waywardmonkeys: Leave and then think about it later.

waywardmonkeys:
Okay, so if you push with the methods on `Sort` removed, I'll merge and then we can iterate. Doctests can definitely come after the merge in this case.

z3/src/context.rs
cdisselkoen: If we leave these in, should they just return `Dynamic` for now?

z3/src/solver.rs
cdisselkoen: I interpret "as is" as "as `-> impl Ast`" for now, so that's in the last force-push. I think `-> impl Ast` is probably a better analogue to the Python `ExprRef` type than `Dynamic` is, anyway.

z3/src/context.rs
waywardmonkeys: Oh, right, these are commented out. Removing them is fine then.

z3/src/context.rs
cdisselkoen: These turn out to be used a bunch even in the existing tests, so I'll just deprecate.

z3/src/context.rs
cdisselkoen: Removed in the latest push

cdisselkoen:
Pushed, and I think that's all of the issues that need to be addressed before merge.  I'm 👍 to merge

waywardmonkeys:
Thanks!

PR #33, opened 2019-06-28
Simplify Symbols.

waywardmonkeys:
Instead of having to pass a context reference and create an
actual symbol, just pass the data for the symbol and let the
function that needs a symbol create it using Symbol::as_z3_symbol().

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -261,11 +261,11 @@ impl<'ctx> Ast<'ctx> for Set<'ctx> {
 }
 
 impl<'ctx> Bool<'ctx> {
-    pub fn new_const(sym: &Symbol<'ctx>) -> Bool<'ctx> {
-        let sort = Sort::bool(sym.ctx);
-        Self::new(sym.ctx, unsafe {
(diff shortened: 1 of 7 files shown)


waywardmonkeys:
@cdisselkoen This is something that I am thinking about ... I would also do the same sort of thing for `Sort` ... it would remove a lot of boilerplate code.

waywardmonkeys:
This happened because I wrote this:

```rust
let f = FuncDecl::new(
    &ctx,
    Symbol::from_string(&ctx, "f"),
    &[&Sort::int(&ctx), &Sort::real(&ctx)],
    &Sort::int(&ctx));
```

And I'd rather get it to look like:

```rust
let f = FuncDecl::new(&ctx, "f", &[Sort::Int, Sort::Real], Sort::Int);
```

waywardmonkeys:
I'm thinking of just killing the methods on `Context` ...

cdisselkoen:
That's good with me too.

waywardmonkeys:
I'll merge this now and deal with `Sort` in the morning (I'm in GMT+7, so it is almost 1am here.).

waywardmonkeys:
Thanks for the feedback, I was looking for it. :)

PR #37, opened 2019-06-30
remove Z3_MUTEX

jrakow:
All bindings types contain raw pointers. Types with raw pointers are not Send (https://doc.rust-lang.org/nomicon/send-and-sync.html). So bindings types cannot be send to other threads. Hence we don't need a global lock.

### DIFF
--- a/z3/Cargo.toml
+++ b/z3/Cargo.toml
@@ -14,7 +14,6 @@ repository = "https://github.com/prove-rs/z3.rs.git"
 
 [dependencies]
 log = "0.4"
-lazy_static = "1"
 
 [dev-dependencies]
 env_logger = "0.6"
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -7,7 +7,6 @@ use z3_sys::*;
 use Context;
 use Sort;
 use Symbol;
-use Z3_MUTEX;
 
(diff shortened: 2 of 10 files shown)


waywardmonkeys:
The notes from the early versions of this crate (before me) said that it was because of issues on the Z3 side during initialization ... perhaps that's been fixed there though. Have you run with these changes extensively?

jrakow:
I have used Z3 in concurrent tests and I never had a single issue. As far as I know, Context, Solver, Model and similar structs are thread-safe. It's just initialization that's hazardous. I would treat that as an upstream bug though.
Locking every single operation seems overkill to me. It makes `cargo test` run at one eighth the speed when you have eight cores. If we were to keep the lock, then we should change it to be only used on Config and other global stuff.

waywardmonkeys:
Okay, let's give it a try!

waywardmonkeys:
So, I suspect that you're running on Windows or Linux.

Up until shortly after the 4.8.5 release, Z3 used OpenMP, which wasn't present on macOS builds, so it was all no-ops.

`master` branch of Z3 has correct locking on all platforms, but that isn't released yet.

As such, that results in the tests crashing on macOS now inside the symbol table code when run against Z3 4.8.5 or earlier (all public releases).

PR #38, opened 2019-07-01
Fix tests

jrakow:
This makes ast::Dynamic not an enum, but just another Ast variant. When it was an enum, the Z3 enumeration type could not be represented as there was no Ast variant. Dynamic is now used both when types are not known at compile time and also when there is now Ast variant like for enumeration. During that I also moved more code into macros to considerably reduce duplication and code size.

There is a caveat: Z3 has no `Z3_sort_kind` for set sorts, so dynamic conversion to set sorts is not possible. I am not sure how to work around that. This is probably an artefact of how Z3 represents sets.

The second commit simply adds a cast in a test case to make it work.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -45,41 +45,10 @@ pub struct Set<'ctx> {
     pub(crate) z3_ast: Z3_ast,
 }
 
-// These `From` impls imply the corresponding `Into` impls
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
LGTM! Thanks! (FYI: @cdisselkoen)

PR #43, opened 2019-07-01
Add datatype variant of Ast and a builder

jrakow:
This creates a way to construct and use Z3 datatypes, which are like Rust enums. There is no support yet for recursive datatypes. Also, accesses are by index and hence a bit unwieldy, but it works.
I don't see a way to emulate Python's nice API in Rust, maybe we can improve on this in the future.


waywardmonkeys:
Thank you!

PR #46, opened 2019-07-24
add more links to the top-level README

cdisselkoen:
It seemed like a good idea to me to have links here to the `z3` and `z3-sys` readmes.

Unrelatedly - what remains on the todo list before we get the next crates.io release of `z3`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,5 @@
-This repository contains high level and low level Rust bindings
+This repository contains [high level] and [low level] Rust bindings
 for the [Z3 solver](https://github.com/Z3Prover/z3).
+
+[high level]: https://github.com/prove-rs/z3.rs/tree/master/z3
+[low level]: https://github.com/prove-rs/z3.rs/tree/master/z3-sys


waywardmonkeys:
For the next release ... I'd like to get the change to make `Sort` act like `Symbol` without having to have a context to create it ... but I am traveling right now until the end of the month ... 

But I guess I can make the release without that and try to get to it soon in August! Unless you'd like to do it?

cdisselkoen:
If I'm interpreting you correctly, you're proposing that `Sort` should be a native enum with no `z3-sys` members and just a function `as_z3_sort` which constructs the Z3 sort. This sounds great to me.

Unfortunately, we can't do this quite exactly how it's currently done for `Symbol`, because unlike Z3 symbols, Z3 sorts are refcounted.  So on first glance I think we'd need two datatypes: the proposed `Sort` enum with no `z3-sys` members, and a `Z3Sort` struct (can bikeshed the name) which looks more like the current `Sort`, i.e., a safe wrapper around the `Z3_Sort` and `Context` which handles the refcounting, has the proper `Drop` trait, etc.  Does that sound like a reasonable approach to you?

cdisselkoen:
Actually, the `Z3Sort` struct I mention above would probably be identical to the current `Sort`, just renamed

waywardmonkeys:
I'll do this and land it within the next day or three.

PR #47, opened 2019-07-24
Add solver parameters

fpoli:
I needed to set a parameter of the solver, so I added the `Params` structure and a `set_params` method to `Solver`.

Please double check the mutex and `unsafe` blocks before merging, because it's the first time that I use `z3_sys`.


waywardmonkeys:
Thanks! This looks good to me. Sorry about the delay. I have been traveling this month.

PR #48, opened 2019-07-26
implemented substitutions for Ast

hengchu:
This PR adds a `substitute` method on the `Ast` trait with a default implementation. It just uses the `Z3_substitute` binding from `z3-sys` to do the actual substitution.

A unit test case was also added :) I will be happy to make additional changes based on maintainers' opinion.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -183,6 +183,38 @@ pub trait Ast<'ctx>: Sized {
             Z3_simplify(self.get_ctx().z3_ctx, self.get_z3_ast())
         })
     }
+
+    /// Performs substitution on the `Ast`. The slice `substitutions` contains a
+    /// list of pairs with a "from" `Ast` that will be substituted by a "to" `Ast`.
+    fn substitute<F, T>(&self, substitutions: &[(&F, &T)]) -> Self
+    where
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
Thanks!

hengchu:
Just wondering: is there a plan of making a new release soon? The current `Ast` design seems quite different from the latest release `0.3.2`.

waywardmonkeys:
@hengchu Soon! I've been traveling for most of this month. I want to go over things and think about some stuff. The new API is much better, but I feel like it could improve a bit more still to simplify things.

How do you feel about the current `Ast` design?

hengchu:
@waywardmonkeys I love it! I think it's much better than the one in `0.3.2` where there's only 1 `Ast` type. The current version is more like the python bindings, but with all the Rust goodies :) Great work!

hengchu:
Although I think it might be worth doing some research on figuring out exactly how thread-safe z3 is. Currently `Context` is `!Send` and `!Sync`. Although we know we can't use context from multiple threads concurrently, I wonder if it's safe to create context in one thread but then completely move it to another thread. It's not a critical use case for me personally, but I think people might find it useful.

PR #49, opened 2019-07-29
adding comparison operators on reals

hengchu:
With a unit test case :)


waywardmonkeys:
Thanks!

I'll land this one for now, but I think I want to look at changing how we implement all of these operators to use some traits and then impl the traits for various types, probably with some output associated types like the core Rust ops traits do. This would let us handle some stuff in a nicer way than we do now ...

PR #54, opened 2019-08-03
implemented functions for building arbitrary size int and reals

hengchu:
with test cases for both


hengchu:
The second commit adds an optional feature that uses the `num` crate's `BigInt` and `BigRational` types for a better interface for constructing these large numerals. If `num` is too much of a dependency to pull in, we can perhaps choose to only add the stringly-typed interface?

Issue #55, opened 2019-08-05
Distinguish "unknown" from "unsat"


fpoli:
Currently, using the `Solver::check` method, there is no way to distinguish an "unknown" answer from an "unsat".

One (non backward compatible) solution would be to:
1. add a `LiftedBool` enum, with variants `True`, `False`, and `Undefined`
2. rename `Solver::check` to `Solver::is_sat`
3. add a `Solver::check` method that returns a `LiftedBool`

waywardmonkeys:
Yes, adding a `LiftedBool` and then changing `Solver::check` and `Solver::check_assumptions` to return that instead of `bool` sounds like the correct thing to do.

If you'd like to do that and submit a PR, let me know. Otherwise, I'll get to it in about 15 hours.

waywardmonkeys:
(and the doc comments will need updating as well.)

waywardmonkeys:
Fixed in #56.

PR #56, opened 2019-08-05
Add SatResult type

fpoli:
See issue #55.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -592,7 +592,7 @@ impl<'ctx> Int<'ctx> {
     ///
     /// # Examples
     /// ```
-    /// # use z3::{ast, Config, Context, Solver};
+    /// # use z3::{ast, Config, Context, SatResult, Solver};
     /// # use z3::ast::Ast;
     /// # let cfg = Config::new();
     /// # let ctx = Context::new(&cfg);
@@ -602,7 +602,7 @@ impl<'ctx> Int<'ctx> {
     ///
     /// let x = ast::Int::from_bv(&bv, true);
     ///
-    /// assert!(solver.check());
+    /// assert_eq!(solver.check(), SatResult::Sat);
     /// let model = solver.get_model();
     ///
     /// assert_eq!(-3, model.eval(&x).unwrap().as_i64().unwrap());
@@ -812,7 +812,7 @@ impl<'ctx> BV<'ctx> {
     ///
     /// # Examples
     /// ```
-    /// # use z3::{ast, Config, Context, Solver};
+    /// # use z3::{ast, Config, Context, SatResult, Solver};
     /// # use z3::ast::Ast;
     /// # let cfg = Config::new();
     /// # let ctx = Context::new(&cfg);
@@ -823,7 +823,7 @@ impl<'ctx> BV<'ctx> {
     /// let x = ast::BV::from_int(&i, 64);
     /// assert_eq!(64, x.get_size());
     ///
-    /// assert!(solver.check());
+    /// assert_eq!(solver.check(), SatResult::Sat);
     /// let model = solver.get_model();
     ///
     /// assert_eq!(-3, model.eval(&x.to_int(true)).unwrap().as_i64().expect("as_i64() shouldn't fail"));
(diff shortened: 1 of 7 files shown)


z3/src/solver.rs
```diff
@@ -213,10 +222,17 @@ impl<'ctx> Solver<'ctx> {
         })
     }
 
-    /// Set the current solver using the given parameters.
-    pub fn set_params(&self, params: &Params<'ctx>) {
-        let guard = Z3_MUTEX.lock().unwrap();
-        unsafe { Z3_solver_set_params(self.ctx.z3_ctx, self.z3_slv, params.z3_params) };
```
waywardmonkeys: I don't think we should remove `set_params` here.

z3/src/solver.rs
```diff
@@ -213,10 +222,17 @@ impl<'ctx> Solver<'ctx> {
...
+    /// Return a brief justification for an "unknown" result (i.e., `LiftedBool::Undefined`) for
+    /// the commands [`Solver::check()`](#method.check) and
+    /// [`Solver::check_assumptions()`](#method.check_assumptions)
+    pub fn get_reason_unknown(&self) -> String {
+        let p = unsafe {
+            CStr::from_ptr(Z3_solver_get_reason_unknown(self.ctx.z3_ctx, self.z3_slv) as *mut i8)
```
waywardmonkeys: I don't think you should need the `as *mut i8` here.

z3/src/solver.rs
fpoli: Oh, I messed up with the rebase. I can fix it tomorrow.

cdisselkoen:
I'm wondering whether it might be preferable to use `Option<bool>` instead of this `LiftedBool` type?  I'm not sure which one I prefer myself, but I was wondering if `Option<bool>` was considered as an option.

waywardmonkeys:
Another might be `Result<bool, String>` where the `bool` is for sat / unsat and the `String` error case is for unknown.

cdisselkoen:
Or leave `LiftedBool` as-is (maybe rename to something like `SATResult`) and add an `impl TryFrom<LiftedBool> for bool`, which would allow callers to get a `Result<bool, ...>` if they wanted one.

fpoli:
Thanks for the suggestions. So, some alternatives are:

1. `Option<bool>`: quick and easy to use.
2. `Result<bool, String>`: "unknown" is considered an error.
3. `LiftedBool::True`/`False`/`Undefined`: a direct translation of `z3_sys::Z3_lbool`.
4. `SatResult::Sat`/`Unsat`/`Unknown`: like `LiftedBool`, but with more meaningful names.
5. A combination of `Result<bool, String>` and `SatResult`, with `impl TryFrom<LiftedBool> for bool`.

I personally like options 1 and 4, but I don't have a strong preference.

waywardmonkeys:
I like #4 as well. This should happen for the `Optimizer` as well.

z3/src/ast.rs
```diff
@@ -592,7 +592,7 @@ impl<'ctx> Int<'ctx> {
     ///
     /// # Examples
     /// ```
-    /// # use z3::{ast, Config, Context, Solver};
+    /// # use z3::{ast, Config, Context, Solver, LiftedBool};
```
waywardmonkeys: Please alphabetize these. (Or just run `cargo fmt`)

z3/src/lib.rs
```diff
@@ -175,3 +175,10 @@ pub struct Params<'ctx> {
     ctx: &'ctx Context,
     z3_params: Z3_params,
 }
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum LiftedBool {
```
waywardmonkeys: After renaming this to `SATResult` or something ... please also add a doc comment here explaining what it is for.

z3/src/solver.rs
```diff
@@ -213,6 +223,19 @@ impl<'ctx> Solver<'ctx> {
         })
     }
 
+    /// Return a brief justification for an "unknown" result (i.e., `LiftedBool::Undefined`) for
+    /// the commands [`Solver::check()`](#method.check) and
+    /// [`Solver::check_assumptions()`](#method.check_assumptions)
+    pub fn get_reason_unknown(&self) -> String {
```
waywardmonkeys: Please use the same code structure as we do in the various `Display` impls so that it matches those in terms of the creation of the string, checking for NULL, etc.

fpoli: I opted to return an `Option<String>`.

waywardmonkeys:
Thanks! This is a great set of improvements!

PR #57, opened 2019-08-05
Add patterns for quantifiers

fpoli:


z3/src/ast.rs
```diff
@@ -642,6 +643,7 @@ impl<'ctx> Int<'ctx> {
     binop!(le, Z3_mk_le, Bool<'ctx>);
     binop!(gt, Z3_mk_gt, Bool<'ctx>);
     binop!(ge, Z3_mk_ge, Bool<'ctx>);
+    binop!(_eq, Z3_mk_eq, Bool<'ctx>);
```
cdisselkoen: The `_eq` method is already defined on the `Ast` trait - is there a reason it needs to be separately defined on `Int`?  I believe this implementation is the same one that's already provided by `Ast`.

z3/src/params.rs
```diff
@@ -62,4 +62,4 @@ impl<'ctx> Drop for Params<'ctx> {
         let guard = Z3_MUTEX.lock().unwrap();
         unsafe { Z3_params_dec_ref(self.ctx.z3_ctx, self.z3_params) };
     }
-}
+}
```
cdisselkoen: I think this change was accidental (it's the only change in this file)

z3/src/ast.rs
fpoli: No, I overlooked the definition in `Ast`.

z3/src/ast.rs
```diff
@@ -1237,38 +1238,97 @@ impl<'ctx> Datatype<'ctx> {
...
 ) -> Dynamic<'ctx> {
-    assert!(bounds.iter().all(|a| a.get_ctx().z3_ctx == ctx.z3_ctx));
+    assert!(bounds.iter().all(|a| a.get_ctx() == ctx));
+    assert!(patterns.iter().all(|p| p.ctx == ctx));
     assert_eq!(ctx, body.get_ctx());
 
-    if bounds.is_empty() {
-        return body.clone();
-    }
-
```
fpoli: I removed this because I would prefer to avoid adding optimizations to this crate (and because Z3 should be able to easily optimize this case). Let me know if you have a different vision.

waywardmonkeys: It might be good to make sure that Z3 actually does optimize this ... Either way, if this is removed, I'd rather see it happen in a separate PR.

waywardmonkeys:
This needs rebasing forward now that #56 has landed.

PR #58, opened 2019-08-06
Derive PartialEq, Eq, etc on Context and Symbol

cdisselkoen:
Cleans up some equality checks in ast.rs, and also allows external crate users to check whether `Context`s are equal (currently only in-crate users are able to do this).  Also adds some other potentially useful traits that can be straightforwardly derived.


waywardmonkeys:
Thanks. I had a local set of changes that I was doing that needed `PartialEq` and such for `Symbol`, so this separates that out from that. :)

PR #60, opened 2019-08-06
Factor out SafeAstPtr and allow Ast trait objects

djrenren:
This commit is largely a refactor that centralizes the logic for Ast pointers into a `SafeAstPtr` type, rather than duplicating it across every AST struct (`Bool`, `Int`, etc.).

This allows us to rely on more Rust machinery (auto-derives) and makes safe vs. unsafe access to the z3 internals more explicit. It also removes the raw `Z3_ast` ptr from the public interface of `Ast` by allowing `Ast::new` to accept a `SafeAstPtr` rather than a raw `Z3_ast` ptr.

While I was at it, I moved the `Sized` restrictions onto the trait methods that required it, allowing trait objects a la `&dyn Ast` to exist. These could replace the methods where we have `& impl Ast` but I figured that's a task for another day.

If people think this new ptr is worthwhile I think there's a way to unify all of the various pointer types so they are automatically safe, but again I think that's best left to a future commit.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -122,7 +122,7 @@ pub struct Datatype<'ctx>(SafeAstPtr<'ctx>);
 #[derive(Hash, Debug, PartialEq, Eq, Clone)]
 pub struct Dynamic<'ctx>(SafeAstPtr<'ctx>);
 
-macro_rules! for_each_static_typed_ast {
+macro_rules! for_each_ast {
     ($m:ident) => {
         $m!(Bool);
         $m!(Int);
@@ -131,12 +131,6 @@ macro_rules! for_each_static_typed_ast {
         $m!(Array);
         $m!(Set);
         $m!(Datatype);
-    };
-}
-
-macro_rules! for_each_ast {
-    ($m:ident) => {
-        for_each_static_typed_ast!($m);
         $m!(Dynamic);
     };
 }
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -1,3 +1,4 @@
+use std::borrow::Borrow;
 use std::cmp::{Eq, PartialEq};
 use std::convert::{TryFrom, TryInto};
 use std::ffi::{CStr, CString};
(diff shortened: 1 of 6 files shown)


waywardmonkeys:
Thanks for this PR. I've been thinking about it today and am still thinking about it.

I understand some of the motivation behind it. I've been thinking about how to address some of that as well.

I need a bit more time!

waywardmonkeys:
I have been thinking about adding an impl of `Deref` though to remove some of the boilerplate, so that you can pass a `Context` and let the `Deref` handle pulling out the `z3_ctx` member.

Also, I have been thinking about renaming a bunch of the `new` methods to be `wrap` when that's what they really are (like the ones for `Ast` implementations). With that, if we did the `SafeAstPtr` thing, then that could readily happen within the `wrap` method and reduce the amount of change...

Thoughts?

djrenren:
> I need a bit more time!

No worries, I kinda came out of nowhere this PR so that's understandable.

>I have been thinking about adding an impl of Deref though to remove some of the boilerplate, so that you can pass a Context and let the Deref handle pulling out the z3_ctx member.

My concern with implementing any `Deref`s to get `Z3_*` types is that this would be exposed to any and all library consumers. Because the `Z3_*` pointers are `Copy` then it's really easy to violate the safety provided by `SafeAstPtr` like so:

```
let uhoh = {
     let ctx: Context<'_> = //...
     *ctx
}
```

When `ctx` gets dropped, we'd call `Z3_del_context`, yet we'd still have the `Z3_ctx` pointer to it lying around. 

I'd much prefer a bit of internal boilerplate to unsafety in the presumably safe API. That said, we could probably cut down on the cruft a tad. For example, I removed `get_context()` from `Ast` but we could add it back in. (`get_z3_ast()` would unsafe to continue with though)

> Also, I have been thinking about renaming a bunch of the `new` methods to be `wrap` when that's what they really are (like the ones for Ast implementations).

Yeah that seems fine. For public interfaces I might consider a more descriptive name though. For example `Ast::ascribe_type` (or hopefully something less wordy).

> With that, if we did the SafeAstPtr thing, then that could readily happen within the wrap method and reduce the amount of change...

I'm unsure what you mean by "*that* could readily happen", could you elaborate?

fitzgen:
Closing this due to inactivity / bit rotting. Not saying this isn't something we want, but it will need to be revisited, and I'd like to understand the motivation a bit more as well. Thanks!

djrenren:
Totally fair! I sorta lost track of this PR in the midst of PhD craziness.

I believe the motivation was that this library had widespread unsafe usage of ast pointers which required notifying z3 of refcount increments at several location. The goal was to create an abstraction in one place that would just take care of the problem and also I think I really wanted to remove the Z3_ast pointer from the public interface, so it couldn't be mishandled... I don't think either motivation was particularly pressing though.

fitzgen:
That motivation sounds good to me: `UntypedAst` (or whatever) handling the underlying reference counting operations and the typed AST types handling type safety. This is vaguely reminiscent of `RawVec` handling allocations/capacity and `Vec` handling initialization/length.

If you, or anyone else, ever wants to dust this off and rebase this PR, or recreate it from scratch, I'd be happy to review and help get it merged!

PR #61, opened 2019-08-13
Add an impl TryFrom<SatResult> for bool

cdisselkoen:
This allows callers to quickly turn a `SatResult` into a `Result<bool>` if they prefer working with a `Result<bool>`.  This was briefly discussed in #56 and discarded, but I'd like to reopen this conversation if that's okay.  Is there a downside to including this impl?

### DIFF
--- a/z3/src/lib.rs
+++ b/z3/src/lib.rs
@@ -13,6 +13,7 @@ extern crate z3_sys;
 #[cfg(feature = "arbitrary-size-numeral")]
 extern crate num;
 
+use std::convert::TryFrom;
 use std::ffi::CString;
 use std::sync::Mutex;
 use z3_sys::*;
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
You would still want to pattern match on this ... so I wasn't sure how it would help. I guess with an `unwrap` for testing code, it would be useful?

cdisselkoen:
`Result` provides a number of useful methods - `unwrap`, `and`/`and_then`, `or`/`or_else`, etc; and perhaps most importantly to me, `?` to propagate any solver errors up through the call stack.

This is also not a big deal, as I can implement a `SatResult` -> `Result<bool>` wrapper function myself and call that everywhere instead of `check()`, but I thought it might be useful to others to include here.

cdisselkoen:
I'm just going to close this, it's easy for anyone who wants this to implement a `SatResult -> Result<bool>` themselves.

PR #67, opened 2020-02-16
Feat: Add method get_ctx() to Solver

LEAFERx:
This might be useful when we need to `assert` and make `ast` objects at the same time, like when doing symbolic execution. With this method, we can pass down `solver` as a parameter only but not pass down `context` simultaneously, thus making interfaces cleaner.

### DIFF
--- a/z3/src/solver.rs
+++ b/z3/src/solver.rs
@@ -73,6 +73,11 @@ impl<'ctx> Solver<'ctx> {
         }
     }
 
+    /// Return the context of the solver.
+    pub fn get_ctx(&self) -> &'ctx Context {
+        self.ctx
+    }
+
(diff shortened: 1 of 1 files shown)


z3/src/solver.rs
```diff
@@ -73,6 +73,11 @@ impl<'ctx> Solver<'ctx> {
         }
     }
 
+    /// Return the context of the solver.
+    pub fn get_ctx(&self) -> &'ctx Context {
```
fitzgen:
```suggestion
    pub fn get_context(&self) -> &'ctx Context {
```

I think it is a little cleaner to avoid contractions / shortening words in public APIs within reason.

PR #79, opened 2020-05-29
Add String sort and AST type 

sphw:
This adds bindings to existing Z3 `String` sort, and an AST mapping for it. The code is largely based off the existing bindings. 

Let me know if there are any issues


sphw:
I added some tests and comments. Let me know if there is anything else

Issue #80, opened 2020-06-12
Improve function signature of Bool::pb_eq for better ergonomics


bumblepie:
Currently, calling `Bool::pb_eq` (and similarly `Bool::pb_lt` and `Bool::pb_gt`) looks something like this:
```
let a = ast::Bool::new_const(&ctx, "a");
let b = ast::Bool::new_const(&ctx, "b");
let c = ast::Bool::new_const(&ctx, "c");
solver.assert(&a.pb_eq( &[&b, &c], vec![1, 2, 3], 3));
```
I don't like this for a few reasons:
- It makes it look like there is some order or precedence that makes `a` more important than the other `Bool`s
- There isn't a clear matching between the booleans and the coefficients: the first coefficient is tied to `a`, but at a glance you would think it would be tied to `b` since `a` is outside of the list within the function call. Additionally, if you have many weights, it is unclear at a glance which weight applies to which 
- It's even less ergonomic when applying to a list:
```
solver.assert(&bools[0].pb_eq( &bools[1..], vec![1; 10], 10));
```
I believe a more intuitive function signature would greatly benefit the usability of these functions.

The first change I would propose would be turning these methods into functions by removing the `&self` reference. Based on the current implementation, we would need to pass in a context to replace this. This would change the above examples to:
```rust
solver.assert(Bool::pb_eq(&ctx, &[&a, &b, &c], vec![1, 2, 3], 3));
solver.assert(Bool::pb_eq(&ctx, &bools[..], vec![1; 10], 10));
```
Already we see a minor improvement in that we have removed the first problem of apparent arbitrary precedence. There's also much clearer matching between the coefficients and the `Bool`s themselves now that the `Bool`s are all grouped together too. We could stop here, but I would also like to propose one more change: if we look at the implementations from other languages like [Python](https://stackoverflow.com/questions/43081929/k-out-of-n-constraint-in-z3py), we can see that instead of passing through a list of `Bool`s and their coefficients separately, they are passed through as a single list of tuples containing both the `Bool` and their coefficient. I would propose that this is also adopted into the new function signature for a couple of reasons:
- The coefficients are clearly aligned and matched with their corresponding `Bool`s
[...]

PR #83, opened 2020-06-27
Dev/non selfy fns

bumblepie:
Solves #80


bumblepie:
Note: this introduces pretty wide-sweeping fn signature changes which aren't backwards-compatible, so it'll need a version bump I think

fitzgen:
> Note: this introduces pretty wide-sweeping fn signature changes which aren't backwards-compatible, so it'll need a version bump I think

Yes, the next release will be a breaking bump.

PR #89, opened 2020-08-10
Feature: add translate on model

LEAFERx:
This pull request added `Model::translate` using `Z3_model_translate`. The test was also added.


z3/src/model.rs
```diff
@@ -32,6 +33,18 @@ impl<'ctx> Model<'ctx> {
         }
     }
 
+    pub fn translate<'dest_ctx>(&self, dest: &'dest_ctx Context) -> Model<'dest_ctx> {
```
fitzgen, line 36: Please add a documentation comment for this method.

Issue #90, opened 2020-08-21
Support for recursive datatype


taegyunkim:
Similar to what's implemented in python binding. https://z3prover.github.io/api/html/namespacez3py.html#a2aae8887c59dd8d4c298e6c0c05405d1

fitzgen:
We have bindings to `Z3_mk_datatypes` in `z3-sys`, so we just need to add the high-level API. Are you interested in contributing this?

taegyunkim:
@fitzgen Yes, unless someone started working on this.

fitzgen:
I don't think anyone has, so please go for it! :)

fitzgen:
Fixed in #93

PR #93, opened 2020-09-17
DatatypeBuilder changes to support recursive datatypes

taegyunkim:
Addresses #90

### DIFF
--- a/z3/src/datatype_builder.rs
+++ b/z3/src/datatype_builder.rs
@@ -31,12 +31,16 @@ impl<'ctx> DatatypeBuilder<'ctx> {
 }
 
 pub fn create_datatypes<'ctx>(ds: &[&DatatypeBuilder<'ctx>]) -> Vec<DatatypeSort<'ctx>> {
-    let ctx: &'ctx Context = ds[0].ctx;
     let num = ds.len();
+    assert!(num > 0, "At least one DatatypeBuilder must be specified");
 
-    assert!(num > 0, "input ds empty");
+
+    let ctx: &'ctx Context = ds[0].ctx;
     let mut names: Vec<Z3_symbol> = Vec::with_capacity(num);
+
     let out: Vec<Z3_sort> = Vec::with_capacity(num);
+    // Prevent running `out's destructor so we are in complete control
+    // of the allocation.
     let mut out = mem::ManuallyDrop::new(out);
     let mut clists: Vec<Z3_constructor_list> = Vec::with_capacity(num);
 
@@ -119,15 +123,11 @@ pub fn create_datatypes<'ctx>(ds: &[&DatatypeBuilder<'ctx>]) -> Vec<DatatypeSort
         );
     };
 
-    println!("returned from z3_mk_datatypes");
+    let out_rebuilt: Vec<Z3_sort> = unsafe { Vec::from_raw_parts(out.as_mut_ptr(), num, num) };
 
-    let rebuilt = unsafe { Vec::from_raw_parts(out.as_mut_ptr(), num, num) };
-
-    assert!(rebuilt.len() > 0, "empty rebuilt vec.");
-
-    let mut datatype_sorts: Vec<DatatypeSort<'ctx>> = Vec::with_capacity(rebuilt.len());
-    for i in 0..rebuilt.len() {
-        let s = rebuilt[i];
+    let mut datatype_sorts: Vec<DatatypeSort<'ctx>> = Vec::with_capacity(out_rebuilt.len());
+    for i in 0..out_rebuilt.len() {
+        let s = out_rebuilt[i];
         let d = &ds[i];
         let num_cs = d.constructors.len();
 
--- a/z3/src/datatype_builder.rs
+++ b/z3/src/datatype_builder.rs
@@ -2,125 +2,11 @@ use std::mem;
 use std::{convert::TryInto, ptr::null_mut};
 use z3_sys::*;
 use {
-    Context, DatatypeAccessor, DatatypeBuilder, DatatypeSort, DatatypeVariant, DtypeBuilder,
-    FuncDecl, Sort, Symbol,
+    Context, DatatypeAccessor, DatatypeBuilder, DatatypeSort, DatatypeVariant, FuncDecl, Sort,
+    Symbol,
 };
 
 impl<'ctx> DatatypeBuilder<'ctx> {
-    pub fn new(ctx: &'ctx Context) -> Self {
-        Self {
-            ctx,
-            variants: Vec::new(),
-        }
-    }
-
-    pub fn variant(mut self, name: &str, fields: &[(&str, &Sort)]) -> Self {
-        let recognizer_name_sym = Symbol::String(format!("is-{}", name)).as_z3_symbol(self.ctx);
-        let name_sym = Symbol::String(name.to_string()).as_z3_symbol(self.ctx);
-
-        assert!(fields
-            .iter()
-            .all(|(name, sort)| sort.ctx.z3_ctx == self.ctx.z3_ctx));
-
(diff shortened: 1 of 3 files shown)


z3/src/datatype_builder.rs
```diff
@@ -1,117 +1,179 @@
...
-            .iter()
-            .all(|(name, sort)| sort.ctx.z3_ctx == self.ctx.z3_ctx));
-
-        let mut field_names: Vec<Z3_symbol> = Vec::with_capacity(fields.len());
-        let mut field_sorts = Vec::with_capacity(fields.len());
-
-        for (name, sort) in fields {
-            field_names.push(Symbol::String(name.to_string()).as_z3_symbol(self.ctx));
-            field_sorts.push(sort.z3_sort);
+    pub fn variant(&mut self, name: &str, fields: &'ctx [(&str, DatatypeAccessor<'ctx>)]) {
```
fitzgen, line 18: Is there a reason this can't continue to be chainable by returning `&mut Self`?

z3/src/datatype_builder.rs
```diff
@@ -1,117 +1,179 @@
...
-                field_sorts.as_ptr(),
-                sort_refs.as_mut_ptr(),
-            )
-        };
+        self.constructors.push((name.to_string(), accessor_vec));
+    }
 
-        self.variants.push((fields.len(), constructor));
-        self
+    pub fn finish(&self) -> DatatypeSort<'ctx> {
```
fitzgen, line 27: This should still take `self`, since attempting to call `finish` multiple times is an error right? (I assume Z3 crashes at an assertion or something)

z3/src/datatype_builder.rs
```diff
@@ -1,117 +1,179 @@
...
-                self.ctx.z3_ctx,
-                name_sym,
-                constructors.len().try_into().unwrap(),
-                constructors.as_mut_ptr(),
-            );
-            Z3_inc_ref(self.ctx.z3_ctx, Z3_sort_to_ast(self.ctx.z3_ctx, s));
-            Sort {
-                ctx: self.ctx,
-                z3_sort: s,
+pub fn create_datatypes<'ctx>(ds: &[&DatatypeBuilder<'ctx>]) -> Vec<DatatypeSort<'ctx>> {
```
fitzgen, line 33: Does this need to be public? I guess if you have multiple data types that each reference each other?

Let's rescope this to a static method of `DatatypeBuilder`: `DatatypeBuilder::create_datatypes`.

Lets also add a smoke test for mutually recursive data types (which was the whole point of this PR, right?).

z3/src/datatype_builder.rs
```diff
@@ -1,117 +1,179 @@
...
 
-        let variants = variants
-            .into_iter()
-            .map(|(num_fields, constructor)| {
-                let mut constructor_func: Z3_func_decl = null_mut();
-                let mut tester: Z3_func_decl = null_mut();
-                let mut accessors: Vec<Z3_func_decl> = Vec::new();
-                accessors.resize(num_fields, null_mut());
+            let constructor = unsafe {
+                Z3_mk_constructor(
```
fitzgen, line 92: We should `let _guard = Z3_MUTEX.lock();` inside the scope of each raw `z3_sys` function call.

Be careful to introduce small block scopes, if necessary, to avoid holding the lock too long or across other higher level library calls that internally try to take the lock and trigger a dead lock (which is annoying to debug).

z3/src/datatype_builder.rs
```diff
@@ -1,117 +1,179 @@
...
+            };
+            cs.push(constructor);
+        }
+        assert!(cs.len() > 0, "Empty cs vec");
 
-                unsafe {
-                    // fill fields
-                    Z3_query_constructor(
+        let clist = unsafe {
+            Z3_mk_constructor_list(ctx.z3_ctx, num_cs.try_into().unwrap(), cs.as_mut_ptr())
```
fitzgen, line 107: Same here.

z3/src/datatype_builder.rs
```diff
@@ -1,117 +1,179 @@
...
+        };
+        clists.push(clist);
+    }
+    assert!(clists.len() > 0, "Empty clist vec");
+
+    assert!(num == names.len());
+    assert!(num == clists.len());
+    unsafe {
+        Z3_mk_datatypes(
```
fitzgen, line 116: And same here.

z3/src/datatype_builder.rs
```diff
@@ -1,117 +1,179 @@
...
+
+    let out_rebuilt: Vec<Z3_sort> = unsafe { Vec::from_raw_parts(out.as_mut_ptr(), num, num) };
+
+    let mut datatype_sorts: Vec<DatatypeSort<'ctx>> = Vec::with_capacity(out_rebuilt.len());
+    for i in 0..out_rebuilt.len() {
+        let s = out_rebuilt[i];
+        let d = &ds[i];
+        let num_cs = d.constructors.len();
+
+        unsafe { Z3_inc_ref(ctx.z3_ctx, Z3_sort_to_ast(ctx.z3_ctx, s)) };
```
fitzgen, line 133: And here.

z3/src/datatype_builder.rs
```diff
@@ -1,117 +1,179 @@
...
+        let sort = Sort { ctx, z3_sort: s };
+
+        let mut variants: Vec<DatatypeVariant<'ctx>> = Vec::with_capacity(num_cs as usize);
+
+        for (j, (cname, fs)) in d.constructors.iter().enumerate() {
+            let num_fs = fs.len();
+            let constructor: FuncDecl<'ctx> = unsafe {
+                let f: Z3_func_decl =
+                    Z3_get_datatype_sort_constructor(ctx.z3_ctx, s, j.try_into().unwrap());
```
fitzgen, line 142: And here.

z3/src/datatype_builder.rs
```diff
@@ -1,117 +1,179 @@
...
+                FuncDecl::from_raw(ctx, f)
+            };
+            let tester = unsafe {
+                let f: Z3_func_decl =
+                    Z3_get_datatype_sort_recognizer(ctx.z3_ctx, s, j.try_into().unwrap());
```
fitzgen, line 147: And here.

z3/src/datatype_builder.rs
```diff
@@ -1,117 +1,179 @@
...
+                FuncDecl::from_raw(ctx, f)
+            };
+            let mut accessors: Vec<FuncDecl<'ctx>> = Vec::new();
+            for k in 0..num_fs {
+                accessors.push(unsafe {
+                    let f: Z3_func_decl = Z3_get_datatype_sort_constructor_accessor(
```
fitzgen, line 153: And here.

fitzgen (commented):
Looks great! Thanks for picking this up @taegyunkim.

A couple small suggestions inline below. Once these are addressed, this PR should be good to merge.

z3/src/datatype_builder.rs
taegyunkim: Done.

z3/src/datatype_builder.rs
taegyunkim: Yes, to supoprt mutually recursive datatypes. 

Self-recursive datatype test is in tests/lib.rs: https://github.com/prove-rs/z3.rs/blob/9ef0ad8a6b2b45a567d7a8628ac5346d2b508232/z3/tests/lib.rs#L577

z3/src/datatype_builder.rs
taegyunkim: Right, good catch.

z3/src/datatype_builder.rs
taegyunkim: No, not really.

fitzgen (approved):
Thanks @taegyunkim !

PR #95, opened 2020-10-12
z3-sys: update z3 to v4.8.8

Philipp15b:
The currently bundled z3 version is not an official release and has some bugs in the arith solver, so I upgraded to v4.8.8.


fitzgen:
Thanks for making this PR. Why not to 4.8.9, the newest release?

fitzgen:
I've opened https://github.com/prove-rs/z3.rs/pull/99 to update us to the latest version of Z3, 4.8.9.

PR #97, opened 2020-10-13
lib is not prepended on windows automatically, so to dynamically link…

### LINKED ISSUE #96: Dynamic linking doesn't work on windows
Your code expects a z3.lib and z3.dll:

https://github.com/prove-rs/z3.rs/blob/3c6ab61482c64b083e79e6ccf550a5fb1e87b0a9/z3-sys/src/lib.rs#L1562-L1563

But building z3 outputs a libz3.lib and libz3.dll.

I checked how debian [packs libz3](https://packages.debian.org/de/sid/libz3-dev), and they ship a libz3.so, not a z3.so. Why do you try to link against z3 instead of libz3, and on what platforms does that work?

Trolldemorted:
… a z3 install we must have an explicit libz3 link

Should fix #96, I tested it locally with a dynamically linked, self built z3

### DIFF
--- a/z3-sys/build.rs
+++ b/z3-sys/build.rs
@@ -53,18 +53,6 @@ fn build_z3() {
 
     let lib = dst.join("lib");
 
-    // For some reason Z3 builds as `libz3.lib`, but on windows the "lib" prefix
-    // is not a convention.
-    if cfg!(target_os = "windows") {
-        let from = lib.join("libz3.lib");
-        let to = lib.join("z3.lib");
-        std::fs::rename(&from, &to).expect(&format!(
-            "failed to rename `{}` to `{}`",
-            from.display(),
-            to.display()
-        ));
-    }
-
     println!("cargo:rustc-link-search=native={}", lib.display());
     println!("cargo:rustc-link-lib=static=z3");
 }
--- a/z3-sys/build.rs
+++ b/z3-sys/build.rs
@@ -54,5 +54,9 @@ fn build_z3() {
     let lib = dst.join("lib");
 
     println!("cargo:rustc-link-search=native={}", lib.display());
-    println!("cargo:rustc-link-lib=static=z3");
+    if cfg!(target_os = "windows") {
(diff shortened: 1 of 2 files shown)


Trolldemorted:
Now it works on windows both statically and dynamically

z3-sys/build.rs
```diff
@@ -53,18 +53,10 @@ fn build_z3() {
...
-        let from = lib.join("libz3.lib");
-        let to = lib.join("z3.lib");
-        std::fs::rename(&from, &to).expect(&format!(
-            "failed to rename `{}` to `{}`",
-            from.display(),
-            to.display()
-        ));
+        println!("cargo:rustc-link-lib=static=libz3");
+    } else {
+        println!("cargo:rustc-link-lib=static=z3");
```
fitzgen, line 60: Nice clean up, thanks.

z3-sys/src/lib.rs
```diff
@@ -7482,3 +7481,11 @@ extern "C" {
...
     pub fn Z3_qe_lite(c: Z3_context, vars: Z3_ast_vector, body: Z3_ast) -> Z3_ast;
 }
+
+#[cfg(not(windows))]
+#[link(name = "z3")]
+extern "C" {}
+
+#[cfg(windows)]
+#[link(name = "libz3")]
+extern "C" {}
```
fitzgen, line 7491: Are these still required even though we are printing the cargo linking commands? I would assume that they are not.

fitzgen (commented):
Thanks! One question below before we merge this.

z3-sys/src/lib.rs
Trolldemorted: Is that properly done for dynamic links too? The function where I changed that is guarded by `#[cfg(feature = "static-link-z3")]`

fitzgen: Ah, good catch. You're correct.

PR #101, opened 2020-10-28
z3-sys: Add Fedora's `lib64` path to rustc's lib search path

### LINKED ISSUE #98: Build failures on fedora when building and statically linking our own Z3
@julian-seward1 could you provide the output of running `cargo build -vv --features static-link-z3` from within the `z3-sys` directory here? Thanks!

fitzgen:
Apparently Fedora will use `lib64` instead of `lib` in some circumstances. This commit should fix the `static-link-z3` build for Fedora.

@julian-seward1, could you try pulling this branch and verifying that it fixes the build issues you are seeing? I'd add a CI job, but github only supports ubuntu. Thanks!

Fixes #98

### DIFF
--- a/z3-sys/build.rs
+++ b/z3-sys/build.rs
@@ -54,14 +54,21 @@ fn build_z3() {
     let mut found_lib_dir = false;
     for lib_dir in &[
         "lib",
-        // Fedora builds seem to use `lib64` rather than `lib` in some
-        // circumstances.
+        // Fedora builds seem to use `lib64` rather than `lib` for 64-bit
+        // builds.
         "lib64",
     ] {
-        let lib_dir = dst.join(lib_dir);
-        if lib_dir.exists() {
+        let full_lib_dir = dst.join(lib_dir);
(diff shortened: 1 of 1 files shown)


julian-seward1:
@fitzgen thanks.  That works for me!
This may or may not be relevant (I have no idea how this path stuff works, but ..) might it be wise to ascertain that we're building for a 64-bit target before looking in `lib64/`?  So we can't accidentally pick up the contents of `lib64/`if it exists for some bizarre reason, but we're actually building for a 32-bit target, and `lib/` is, for some other bizarre reason, empty.

fitzgen:
>  This may or may not be relevant (I have no idea how this path stuff works, but ..) might it be wise to ascertain that we're building for a 64-bit target before looking in `lib64/`? So we can't accidentally pick up the contents of `lib64/`if it exists for some bizarre reason, but we're actually building for a 32-bit target, and `lib/` is, for some other bizarre reason, empty.

Done! Thanks for checking!

PR #103, opened 2020-10-30
Improve function signature of distinct for better ergonomics and consistency

0xbe7a:
The same as #80
Currently, the distinct function gets called on the first value with the remaining values as a parameter. This is a very awkward syntax and now even inconsistent compared to other methods


fitzgen (approved):
Thanks!

PR #105, opened 2020-11-27
Add Ast methods for accessing child nodes

### LINKED ISSUE #106: Accessors for Ast children?
We're trying to port a Python lib that uses the Python Z3 bindings to Rust.  One functionality gap we came across is the ability to traverse the Ast by accessing a node's children.  In python, there are [functions](https://github.com/Z3Prover/z3/blob/master/src/api/python/z3/z3.py#L1026) for `children`, `num_args`, and `arg`.  I'd be very interested in those in z3.rs!

I've opened an RFC PR #105.  Some reasons for caution:
1. I don't really understand the Z3 C API and the right memory model to make sure my `unsafe` blocks are actually safe.
2. I chose the more intuitive names of `num_children` and `nth_child` rather than `num_args` and `arg`, but maybe it's better to keep the APIs similar?
3. I've just done a few basic tests.  I'm happy to do more, if you think I'm going in the right direction.

And thanks for making this crate, btw!  It's very helpful to us.

jagill:
To traverse the Ast, one needs access to the (possibly empty) children of each
node.  This commit introduces:
1. num_children(&self) -> usize: The number of children, possibly 0.
2. nth_child(&self, idx) -> Dynamic: The nth child as a Dyanmic node.
3. children(&self) -> Vec<Dynamic>: The children as a Vec.

Fixes #106

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -236,6 +236,40 @@ pub trait Ast<'ctx>: Sized + fmt::Debug {
             )
         })
     }
+
+    /// Return the number of children of this AST node.
+    /// Leaf nodes (eg Bool consts) will return 0.
+    fn num_children(&self) -> usize {
+        let this_ctx = self.get_ctx().z3_ctx;
+        unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            let this_app = Z3_to_app(this_ctx, self.get_z3_ast());
+            Z3_get_app_num_args(this_ctx, this_app) as usize
+        }
+    }
+
+    /// Return the nth child of this AST node.  Out-of-bound indices will
+    /// return none.
+    fn nth_child(&self, idx: usize) -> Option<Dynamic<'ctx>> {
+        if idx >= self.num_children() {
+            None
+        } else {
+            let idx = u32::try_from(idx).unwrap();
+            let this_ctx = self.get_ctx().z3_ctx;
+            let child_ast = unsafe {
+                let guard = Z3_MUTEX.lock().unwrap();
+                let this_app = Z3_to_app(this_ctx, self.get_z3_ast());
+                Z3_get_app_arg(this_ctx, this_app, idx)
+            };
+            Some(Dynamic::new(self.get_ctx(), child_ast))
+        }
+    }
+
+    /// Return the children of this node as a Vec of Dynamic nodes.
+    fn children(&self) -> Vec<Dynamic<'ctx>> {
+        let n = self.num_children();
+        (0..n).map(|i| self.nth_child(i).unwrap()).collect()
+    }
 }
 
 macro_rules! impl_ast {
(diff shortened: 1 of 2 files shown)


z3/src/ast.rs
```diff
@@ -236,6 +236,39 @@ pub trait Ast<'ctx>: Sized + fmt::Debug {
...
+            let this_app = Z3_to_app(this_ctx, self.get_z3_ast());
+            Z3_get_app_num_args(this_ctx, this_app) as usize
+        }
+    }
+
+    /// Return the nth child of this AST node.  Out-of-bound indices will panic.
+    fn nth_child(&self, idx: usize) -> Dynamic<'ctx> {
+        if idx >= self.num_children() {
+            panic!("Index out of bounds: {} max index {}", idx, self.num_children());
+        }
```
fitzgen, line 255: Rather than panicking, let's change the function's signature to returning `Option<Dynamic<'ctx>>` and return `None` here.

fitzgen (commented):
This looks great, thanks! One small suggestion regarding `nth_child` below, and once you've implemented that change, could you also add a test for getting an out-of-bounds child?

jagill:
Ok, comments addressed. I introduced an `unsafe` method to get children without bounds checking, so we don't do unnecessary checks for the `children()` method (for which we already ensure index validity).

Part of me feels I should have more extensive tests for all the various Ast node kinds, but this code was pretty much lifted from the official python bindings, so it seems likely that this works in general.

cc @siegebell

z3/src/ast.rs
```diff
@@ -236,6 +236,47 @@ pub trait Ast<'ctx>: Sized + fmt::Debug {
...
+        (0..n)
+            .map(|i| Dynamic::new(self.get_ctx(), unsafe { nth_child_ast_unchecked(self, i) }))
+            .collect()
+    }
+}
+
+// Retrieve the child at index `idx`.
+// This **does not check bounds**, the caller **must** ensure that
+// `idx < node.num_children()`.
+unsafe fn nth_child_ast_unchecked<'ctx>(node: &impl Ast<'ctx>, idx: usize) -> Z3_ast {
```
fitzgen, line 274: I would prefer the simpler version where this function's body is inlined into `nth_child` and `children` calls `self.nth_child(i).unwrap()` unless there is a *really* good reason not to do so.

I would totally expect LLVM to easily determine that `i < self.num_children()` inside `children` and remove the unwrap checks from `self.nth_child(i).unwrap()`. Are you seeing otherwise? Also, the cost of the (very predictable) unwrap check should be dwarfed by the function call (especially if Z3 is in a shared library).

jagill:
Fixed!  I haven't checked out the LLVM instructions or done any sort of profiling, so I apologize for any premature optimization.  For context, I've been bitten a few times by 30% performance regressions in code for which bounds checks are not elided, which is where my instincts come from.  But I've neither looked for, nor found them, here!

PR #107, opened 2020-11-29
ast: support converting Dynamic back into Set

sameer:
Check that the range of the array is boolean & allow the conversion. Since a set is just an array with a boolean range, it would be possible to convert a boolean-ranged, array-sorted node into a set-sorted one.


fitzgen (approved):
LGTM, thanks. Will wait on the other PR merging before this one (unless you want to rebase).

fitzgen (approved):
Thanks!

fitzgen:
Can you resolve the conflict? I'll merge afterwards.

sameer:
Squashed it into a single commit so it's clearer

fitzgen (approved):
Thanks!

PR #108, opened 2020-11-30
correct bug in membership: Z3_mk_set_add -> Z3_mk_set_member

sameer:
This should be member. Also looks like element comes first for this function: https://docs.rs/z3-sys/0.6.3/z3_sys/fn.Z3_mk_set_member.html


fitzgen (commented):
Thank you! Would you mind adding a smoke test to `z3.rs/z3/tests/lib.rs` that creates an empty set, asserts that something isn't a member, adds the something to the set, and then asserts that it is a member?

sameer:
Sure thing

fitzgen (approved):
Thanks!

PR #109, opened 2020-11-30
support empty set initialization

sameer:
This brings support for sets that are empty by default, which is the suggested implementation for a question on StackOverflow: https://stackoverflow.com/questions/16100616/defining-set-in-z3


fitzgen (approved):
Thanks!

PR #110, opened 2020-12-01
ergonomics: remove necessity of into() where possible

sameer:
Adding type parameters to these methods makes it possible to pass non-dynamic ast nodes. It can't be done for slice-accepting methods since those arguments *do* need to be dynamic.

Unfortunately, this is a breaking change that would require a version upgrade -- previous calls to these methods with an into() will fail to compile:

```
type annotations needed

cannot infer type for type parameter `A` declared on the function `forall_const`

note: cannot satisfy `_: z3::ast::Ast<'_>`
help: consider specifying the type arguments in the function call: `::<'ctx, A>`rustc(E0283)
lib.rs(301, 18): cannot infer type for type parameter `A` declared on the function `forall_const`
lib.rs(301, 70): this method call resolves to `T`
ast.rs(1477, 8): required by this bound in `z3::ast::forall_const`
```

So some discussion is needed on whether there is a better way of doing this, or if it's even worth breaking existing behavior.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -141,7 +141,7 @@ macro_rules! varop {
 
 /// Abstract syntax tree (AST) nodes represent terms, constants, or expressions.
 /// The `Ast` trait contains methods common to all AST subtypes.
-pub trait Ast<'ctx>: Sized + fmt::Debug {
+pub trait Ast<'ctx>: Sized + fmt::Debug + Into<Dynamic<'ctx>> {
     fn get_ctx(&self) -> &'ctx Context;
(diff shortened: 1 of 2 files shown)


fitzgen:
> So some discussion is needed on whether there is a better way of doing this, or if it's even worth breaking existing behavior.

This also avoids a bunch of `Dynamic` object creation (and likely destruction as well) which requires taking the z3 lock and incrementing ref counts via a function call (that is likely in a different shared library). Which adds up to quite a lot.

I think it is worth the small upgrade costs to get better ergonomics and performance here.

fitzgen (approved):
Thanks!

PR #111, opened 2020-12-03
array store: use reference for index and value

sameer:
Realized I missed making these references in #110 

Hope you didn't publish 0.8.0 to crates.io yet, sorry about that.


fitzgen (approved):
Already published 0.8.0 but I'll just publish 0.9.0 right away.

Thanks!

PR #112, opened 2020-12-05
check context for ops, fixes #94

### LINKED ISSUE #94: Stack overflow when mixing Contexts
The following code causes a stack overflow:

```rust
#[test]
fn mix() {
    let cfg = Config::new();
    let ctx0 = Context::new(&cfg);
    let ctx1 = Context::new(&cfg);
    let i0 = ast::Int::fresh_const(&ctx0, "a");
    let i1 = ast::Int::fresh_const(&ctx1, "b");
    ast::Int::add(&ctx0, &[&i0, &i1]);
}
```

```
thread 'mix' has overflowed its stack
```

The C++/Python/C# bindings check to make sure the contexts are the same before calling the Z3 functions. Rust might be able to get away without runtime checking by encoding this into the type system.

sameer:


fitzgen (approved):
Thank you!

PR #113, opened 2020-12-06
accept refs for the sort datatype accessor

sameer:
this makes re-using a particular sort with accessors easier, since it
can just be referenced in the builder:

(psuedocode)
```rust
let some_complex_sort = Sort::set(&ctx, &Sort::set(...));
let datatype_using_that_sort = DatatypeBuilder::new(&ctx, "foo")
.variant("first", vec![("bar", &some_complex_sort)])
.variant("second", vec![("bar", &some_complex_sort), ("count", &Sort::int(&ctx))])
.finish();
```

It is backwards incompatible -- users of the builder would need to add a reference. Also, if the builder is not `finish`-ed immediately, the sort needs a let binding.

### DIFF
--- a/z3/src/datatype_builder.rs
+++ b/z3/src/datatype_builder.rs
@@ -6,7 +6,7 @@ use {
     Symbol,
 };
 
-impl<'ctx> DatatypeBuilder<'ctx> {
+impl<'sort, 'ctx: 'sort> DatatypeBuilder<'sort, 'ctx> {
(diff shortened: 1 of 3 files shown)


fitzgen (approved):
Thanks!

fitzgen:
Do you have any other updates like this incoming that I should wait for, or should I cut a release now?

sameer:
Sorry for the delay, I missed seeing the email notification for this. Looks like there are a few other changes on the way from others. I can help out and review if needed.

PR #114, opened 2020-12-11
Fixed get_proof deadlock

GreenBeard:
I'm not sure exactly of the purpose of `Z3_MUTEX.lock().unwrap();` so I possibly am not locking properly, but there was a deadlock when calling this function (happened 100% of the time).

### DIFF
--- a/z3/src/solver.rs
+++ b/z3/src/solver.rs
@@ -267,14 +267,15 @@ impl<'ctx> Solver<'ctx> {
     // This seems to actually return an Ast with kind `SortKind::Unknown`, which we don't
     // have an Ast subtype for yet.
     pub fn get_proof(&self) -> Option<impl Ast<'ctx>> {
-        let guard = Z3_MUTEX.lock().unwrap();
-        Some(ast::Dynamic::new(self.ctx, unsafe {
-            let m = Z3_solver_get_proof(self.ctx.z3_ctx, self.z3_slv);
-            if m.is_null() {
-                return None;
-            }
-            m
-        }))
+        let m = unsafe {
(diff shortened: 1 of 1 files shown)


z3/src/solver.rs
```diff
@@ -268,13 +268,13 @@ impl<'ctx> Solver<'ctx> {
...
         let guard = Z3_MUTEX.lock().unwrap();
-        Some(ast::Dynamic::new(self.ctx, unsafe {
-            let m = Z3_solver_get_proof(self.ctx.z3_ctx, self.z3_slv);
-            if m.is_null() {
-                return None;
-            }
-            m
-        }))
+        let m = unsafe { Z3_solver_get_proof(self.ctx.z3_ctx, self.z3_slv) };
+        std::mem::drop(guard);
```
fitzgen, line 272: Can you make this into

```rust
let m = unsafe {
    let _guard = Z3_MUTEX.lock().unwrap();
    Z3_solver_get_proof(...)
};
```

fitzgen (commented):
Thanks! One nitpick below before merging this

z3/src/solver.rs
GreenBeard: Can Rust delay dropping a structure if optimizations are on? I was trying to find clarification on the rules, so I was explicit, but if implicit works too I can totally change it.

fitzgen: No, I just find it easier to read the code and follow what is going on with explicit blocks.

GreenBeard: Fixed!

fitzgen (approved):
Thanks!

PR #116, opened 2020-12-11
Expose Ast::decl() method to return FuncDecl

jagill:
In the C API, Ast nodes that are `expr` expose the underlyling `FuncDecl`, which allows introspection of the type and name of the node.

One question is about the check whether the `Ast` is an `app` or not.  It seems like by construction they are, but I don't have full confidence in this, and would love a second opinion!

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -5,6 +5,7 @@ use std::fmt;
 use std::hash::{Hash, Hasher};
 use z3_sys::*;
 use Context;
+use FuncDecl;
 use Pattern;
 use Sort;
 use Symbol;
@@ -299,6 +300,23 @@ pub trait Ast<'ctx>: Sized + fmt::Debug + Into<Dynamic<'ctx>> {
     /// Note that constants are function applications with 0 arguments.
     fn is_const(&self) -> bool {
         self.is_app() && self.num_children() == 0
+    }
+
+    /// Return the `FuncDecl` of the `Ast`.
+    ///
+    /// This will panic if the `Ast` is not an app, i.e. if AstKind is not App
+    /// or Numeral.
+    fn decl(&self) -> FuncDecl<'ctx> {
+        let ctx = self.get_ctx();
+        let func_decl = unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            let app = Z3_to_app(ctx.z3_ctx, self.get_z3_ast());
+            assert!(!app.is_null(), "Ast has a null Z3_app.");
+            let func_decl = Z3_get_app_decl(ctx.z3_ctx, app);
+            assert!(!func_decl.is_null(), "Ast has a null Z3_func_decl.");
+            func_decl
+        };
+        unsafe { FuncDecl::from_raw(ctx, func_decl) }
     }
 }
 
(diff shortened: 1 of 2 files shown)


fitzgen:
Are these issues fixed in #118? Can this PR be closed now?

jagill:
I carved out the safe parts of this PR.  The parts that are failing are the paths that go Ast -> FuncDecl -> accessor.  Eg, 
get errors when getting what DeclKind a given Ast is by the path `ast.func_decl().kind()`.

fitzgen:
@sameer if you have a few cycles, I'd appreciate some investigation into the `DeclKind` issues that @jagill detailed above. No worries if you don't have time though.

jagill:
Ok, I think we've worked through the issues!  The main question is whether we think it's possible for a Rust Ast node to not be a `z3_app` -- if so, the signature should be `Option<FuncDecl>`.

fitzgen (commented):
I don't know whether any of the ASTs aren't considered applications by Z3. (If someone else knows more about this corner of Z3, sharing your knowledge would be appreciated!) That said, I think it is fine to unconditionally return a `FuncDecl`, rather than an `Option<FuncDecl>`, with the inline suggestions below applied.

Thanks!

jagill:
Added asserts.

z3/src/ast.rs
```diff
@@ -300,6 +301,20 @@ pub trait Ast<'ctx>: Sized + fmt::Debug + Into<Dynamic<'ctx>> {
     fn is_const(&self) -> bool {
         self.is_app() && self.num_children() == 0
     }
+
+    /// Return the `FuncDecl` of the `Ast`.
```
sameer, line 305: Could you add a note for the `self.kind() == AstKind::App` precondition?

jagill: Do you mean in the `is_const()` fn (via `self.is_app()`)?  I want to admit I'm mostly just porting from the python/C++ APIs:
```cpp
         bool is_app() const { return kind() == Z3_APP_AST || kind() == Z3_NUMERAL_AST; }
```
and the various Python asserts that an Ast is actually an app:
```python
    def decl(self):
        if z3_debug():
            _z3_assert(is_app(self), "Z3 application expected")
        return FuncDeclRef(Z3_get_app_decl(self.ctx_ref(), self.as_ast()), self.ctx)
```

I suspect that the python assert -- which is only in debug mode! -- is based on python's loose typing, and it may be unnecessary for the Rust library, but I erred on the side of caution.  (It's worth noting that converting from an Z3_ast to Z3_app is just a cast, and the pointer is unchanged.)

In C++, the class hierarchy is:
* `ast`
* `func_decl` subclass of `ast`
* `expr` subclass of `ast`

where `expr` seems most analogous to the Rust `Ast`.

In short, I don't really understand the class hierarchy model, and I'm happy to do whatever others think is safest.

sameer: I meant just to mention that the method will panic if the ast kind is not AstKind::App

jagill: Added that message in the doc comment.

fitzgen (approved):
Thanks!

PR #119, opened 2020-12-18
Add String::as_string() method

jagill:
Other literal types (Int, Bool, etc) have methods like
`as_bool() -> Option<bool>` or `as_i64() -> Option<i64>` allowing
a literal Ast node to be "unwrapped".  `ast::String` has no such
method, and `to_string` escapes quotes/etc, which makes it challenging
to extract the original stored value.

This introduces an analogous `ast::String::as_string` method, which
returns the original string, unchanged.  `String` constants return
None.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -1053,6 +1053,27 @@ impl<'ctx> String<'ctx> {
         }))
     }
 
+    /// Retrieves the underlying `std::string::String`
+    ///
+    /// If this is not a constant `z3::ast::String`, return `None`.
+    ///
+    /// Note that `to_string()` provided by `std::string::ToString` (which uses
+    /// `std::fmt::Display`) returns an escaped string. In contrast,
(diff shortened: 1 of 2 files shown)


z3/src/ast.rs
```diff
@@ -1035,6 +1035,24 @@ impl<'ctx> String<'ctx> {
         }))
     }
 
+    /// Retrieves the underlying string
+    ///
+    /// If this is not a constant String, return None. Note that `to_string()`
```
fitzgen, line 1040: Can you clarify that this is referring to the `ToString` trait's method, which uses the `std::fmt::Display` implementation? Otherwise users/readers might be wondering where the `z3::ast::String::to_string` method is defined.

fitzgen (commented):
Looks good with one nitpick below addressed. Thanks!

jagill:
Updated comment to better disambiguate between `std::string::String` and `z3::ast::String` and their functions.

fitzgen (approved):
Thanks!

PR #120, opened 2020-12-23
Implement Sort::{is_array, array_domain, array_range}

jagill:
These three functions (present in the C++/Python API) allow one to check if a `Sort` is array-like (`Array` or `Set`), and if so to check the `Sort` of the domain and range.  Non-array-like `Sort`s will return `None` for the latter two functions.


fitzgen (approved):
Thanks!

PR #121, opened 2020-12-27
Optimize: add soft asserts and ability to get objective

sameer:
I've added a testcase that asserts on both.

It's a bit large, let me know if I should pull it out into a separate file / come up with something a bit simpler.

### DIFF
--- a/z3/src/optimize.rs
+++ b/z3/src/optimize.rs
@@ -1,13 +1,20 @@
-use ast::{Ast, Bool};
+use ast::{Ast, Bool, Dynamic};
 use std::convert::TryInto;
-use std::ffi::CStr;
+use std::ffi::{CStr, CString};
 use std::fmt;
 use z3_sys::*;
 use Context;
 use Model;
 use Optimize;
 use SatResult;
+use Symbol;
 use Z3_MUTEX;
+
+#[cfg(feature = "arbitrary-size-numeral")]
+use num::{
+    bigint::{BigInt, BigUint, Sign},
+    rational::BigRational,
+};
 
 impl<'ctx> Optimize<'ctx> {
     /// Create a new optimize context.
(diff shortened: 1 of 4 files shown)


z3/src/optimize.rs
```diff
@@ -34,6 +35,37 @@ impl<'ctx> Optimize<'ctx> {
...
+    /// Group is an optional identifier to group soft constraints.
+    ///
+    /// # See also:
+    ///
+    /// - [`Optimize::maximize()`](#method.maximize)
+    /// - [`Optimize::minimize()`](#method.minimize)
+    pub fn assert_soft<W: ToString>(
+        &self,
+        ast: &impl Ast<'ctx>,
+        weight: W,
```
fitzgen, line 49: The `W: ToString` bound seems overly permissive here. Am I understanding correctly that it should be allowed to be either an integer, a float, or some kind of a rational number string like `"1 / 2"`?

I think we can tighten this up so that users can't pass in things like `"blue hippopotamus"` -- resulting in who knows what behavior from Z3 -- by defining a type and a few conversion impls:

```rust
pub enum Weight {
    Int(i64),
    Float(f64),
    Rational(i64, i64),
}

// i64 -> Weight::Int
impl From<i64> for Weight { /* ... */ }

// f64 -> Weight::Float
impl From<f64> for Weight { /* ... */ }

// (i64, i64) -> Weight::Rational
impl From<(i64, i64)> for Weight { /* ... */ }
```

And then instead of accepting any `W: ToString`, this function would accept any `W: Into<Weight>`, do `w.into()` to convert it into a weight, and then match on which kind of weight it is and format it into a C string as expected by Z3.

z3/src/optimize.rs
```diff
@@ -117,6 +149,34 @@ impl<'ctx> Optimize<'ctx> {
...
+    pub fn get_objectives(&self) -> Vec<Dynamic<'ctx>> {
+        let z3_objectives = unsafe {
+            let _guard = Z3_MUTEX.lock().unwrap();
+            Z3_optimize_get_objectives(self.ctx.z3_ctx, self.z3_opt)
+        };
+
+        let len = unsafe {
+            let _guard = Z3_MUTEX.lock().unwrap();
+            Z3_ast_vector_size(self.ctx.z3_ctx, z3_objectives)
+        };
```
fitzgen, line 164: We could consolidate the unsafe blocks and lock acquisitions like:

```suggestion
        let (z3_objectives, len) = unsafe {
            let _guard = Z3_MUTEX.lock().unwrap();
            let objectives = Z3_optimize_get_objectives(self.ctx.z3_ctx, self.z3_opt);
            let len = Z3_ast_vector_size(self.ctx.z3_ctx, objectives);
            (objectives, len)
        };
```

z3/tests/lib.rs
```diff
@@ -829,3 +829,150 @@ fn test_issue_94() {
     let i1 = ast::Int::fresh_const(&ctx1, "b");
     ast::Int::add(&ctx0, &[&i0, &i1]);
 }
+
+#[test]
+/// Given COUNT items with indices 0..COUNT,
+/// find a function that reverses the indices with soft assertions.
+///
+/// An additional, incorrect soft assertion (COUNT-1 > 0) is used to
+/// show that optimization incurred the least penalty.
```
fitzgen, line 838: Can you put this in a new `objectives.rs` test file or something like that? Thanks.

fitzgen (commented):
Thanks for the PR @sameer! Just a couple nitpicks to address before we merge this. See inline comments below.

z3/src/optimize.rs
sameer: Good point, I think that's a good idea. Will revise

z3/tests/lib.rs
sameer: sure thing

z3/src/optimize.rs
sameer: I'm looking at the function and it's unclear even to me, so I'll dig a bit deeper and figure out what exact weights are valid

sameer: I settled on using the sealed trait pattern to implement to_string on all integer types and integer pairs (for rationals)

fitzgen (approved):
Thanks!

PR #123, opened 2021-01-20
Added functions to get significand and exponent of floating point sorts

DavidThien:
As title says. Pretty small change to add some functions I needed.

### DIFF
--- a/z3/src/sort.rs
+++ b/z3/src/sort.rs
@@ -154,7 +154,9 @@ impl<'ctx> Sort<'ctx> {
         unsafe { Z3_get_sort_kind(self.ctx.z3_ctx, self.z3_sort) }
     }
 
-    pub fn get_float_exponent_size(&self) -> Option<u32> {
+    /// Returns `Some(e)` where `e` is the number of exponent bits if the sort
+    /// is a `FloatingPoint` and `None` otherwise.
+    pub fn float_exponent_size(&self) -> Option<u32> {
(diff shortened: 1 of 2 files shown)


z3/src/sort.rs
```diff
@@ -154,6 +154,22 @@ impl<'ctx> Sort<'ctx> {
         unsafe { Z3_get_sort_kind(self.ctx.z3_ctx, self.z3_sort) }
     }
 
+    pub fn get_float_exponent_size(&self) -> Option<u32> {
```
fitzgen, line 157: Please remove the `get_` prefix, as it is generally unidiomatic Rust (this crate isn't fully consistent with this practice, but gating new changes on being idiomatic helps move us in the right direction).

fitzgen (commented):
Looks great! Before I merge this, would you mind adding documentation comments for each method describing what they do, the units of the returned value, when `None` is returned, etc? Thanks!

DavidThien:
Added documentation and updated the function names

fitzgen (commented):
Oh sorry, I forgot to mention this last time, but one more thing: can you make a smoke test in `tests/lib.rs` double checking that we get the expected significand and exponent sizes for a float? And `None` for an integer or some other non-float type?

Thanks!!

DavidThien:
Np! Added a test as well

fitzgen (approved):
Thanks!

Issue #124, opened 2021-01-22
Implement interface for Tactic and Goal


rlkelly:
There is currently no abstraction on Tactic or Goal, which are useful for things like simplification.  I started working on implementing them, but wanted to check if there had been some work on this in the past or any blockers for adding them to the library.

fitzgen:
> wanted to check if there had been some work on this in the past or any blockers for adding them to the library.

None that I'm aware of; go for it!

rlkelly:
to follow up, I made a PR here: https://github.com/prove-rs/z3.rs/pull/125/

PR #126, opened 2021-01-26
Implement Tactic and goal

rlkelly:
This pull request is in reference to #124. It adds an implementation of goal and tactic containing the most useful methods along with tests for those methods.

### DIFF
--- a/z3/src/goal.rs
+++ b/z3/src/goal.rs
@@ -1,0 +1,162 @@
+use std::ffi::CStr;
+use std::fmt;
+
+use z3_sys::*;
+
+use crate::ast;
+use crate::ast::Ast;
+use Context;
+use Goal;
+use Z3_MUTEX;
+
+
+impl<'ctx> Goal<'ctx> {
+    pub fn new_from_z3_type(ctx: &'ctx Context, z3_goal: Z3_goal) -> Goal<'ctx>{
+        Goal {
+            ctx,
+            z3_goal: z3_goal,
+        }
+    }
+
+    pub fn new(ctx: &'ctx Context, models: bool, unsat_cores: bool, proofs: bool) -> Goal<'ctx> {
+        let goal = Self {
+            ctx,
+            z3_goal: unsafe {
+                let guard = Z3_MUTEX.lock().unwrap();
+                let g = Z3_mk_goal(ctx.z3_ctx, models, unsat_cores, proofs);
+                Z3_goal_inc_ref(ctx.z3_ctx, g);
+                g
+            }
+        };
+        goal
+    }
+
+    /// Add a new formula `a` to the given goal.
+    pub fn assert(&self, ast: &impl ast::Ast<'ctx>) {
+        unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            Z3_goal_assert(self.ctx.z3_ctx, self.z3_goal, ast.get_z3_ast())
+        }
+    }
+
+    /// Return true if the given goal contains the formula `false`.
+    pub fn is_inconsistent(&self) -> bool {
+        unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            Z3_goal_inconsistent(self.ctx.z3_ctx, self.z3_goal)
+        }
+    }
+
+    /// Return the depth of the given goal. It tracks how many transformations were applied to it.
+    pub fn get_depth(&self) -> u32 {
+        unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            Z3_goal_depth(self.ctx.z3_ctx, self.z3_goal)
+        }
+    }
+
+    /// Return the number of formulas in the given goal.
+    pub fn get_size(&self) -> u32 {
+        unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            Z3_goal_size(self.ctx.z3_ctx, self.z3_goal)
+        }
+    }
+
+    /// Return the number of formulas, subformulas and terms in the given goal.
+    pub fn get_num_expr(&self) -> u32 {
+        unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            Z3_goal_num_exprs(self.ctx.z3_ctx, self.z3_goal)
+        }
+    }
+
+    /// Return true if the goal is empty, and it is precise or the product of a under approximation.
+    pub fn is_decided_sat(&self) -> bool {
+        unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            Z3_goal_is_decided_sat(self.ctx.z3_ctx, self.z3_goal)
+        }
+    }
+    /// Return true if the goal contains false, and it is precise or the product of an over approximation.
+    pub fn is_decided_unsat(&self) -> bool {
+        unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            Z3_goal_is_decided_unsat(self.ctx.z3_ctx, self.z3_goal)
+        }
+    }
+
+    /// Erase all formulas from the given goal.
+    pub fn reset(&self) {
+        unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            Z3_goal_reset(self.ctx.z3_ctx, self.z3_goal)
+        };
+    }
+
+    /// Copy a goal `g` from the context `source` to the context `target`.
+    pub fn translate<'dest_ctx>(self, ctx: &'dest_ctx Context) -> Goal<'dest_ctx> {
+        Goal {
+            ctx,
+            z3_goal: unsafe {
+                let guard = Z3_MUTEX.lock().unwrap();
+                Z3_goal_translate(self.ctx.z3_ctx, self.z3_goal, ctx.z3_ctx)
+            }
+        }
+    }
+
+    /// Return the "precision" of the given goal. Goals can be transformed using over and under approximations.
+    pub fn get_precision(&self) -> GoalPrec {
+        let guard = Z3_MUTEX.lock().unwrap();
+        unsafe {
+            Z3_goal_precision(self.ctx.z3_ctx, self.z3_goal)
+        }
+    }
+
+    /// Return a vector of the formulas from the given goal.
+    pub fn get_formulas<T>(&self) -> Vec<T> where T: Ast<'ctx> {
+        let goal_size = self.get_size();
+        let formula = unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            Z3_goal_formula(self.ctx.z3_ctx, self.z3_goal, 0)
+        };
+        let mut formulas = vec![T::new(&self.ctx, formula)];
+
+        for i in 1..goal_size {
+            let formula = unsafe {
+                let guard = Z3_MUTEX.lock().unwrap();
+                Z3_goal_formula(self.ctx.z3_ctx, self.z3_goal, i)
+            };
+            formulas.push(T::new(&self.ctx, formula));
+        };
+        formulas
+    }
+}
+
+impl<'ctx> fmt::Display for Goal<'ctx> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        let p = unsafe { Z3_goal_to_string(self.ctx.z3_ctx, self.z3_goal) };
+        if p.is_null() {
+            return Result::Err(fmt::Error);
+        }
+        match unsafe { CStr::from_ptr(p) }.to_str() {
+            Ok(s) => write!(f, "{}", s),
+            Err(_) => Result::Err(fmt::Error),
+        }
+    }
+}
+
+impl<'ctx> fmt::Debug for Goal<'ctx> {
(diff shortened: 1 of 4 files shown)


rlkelly:
@fitzgen how does this look?

fitzgen:
@sameer do you happen to have cycles to take a look at this? I am a bit busy right now, but this PR (and earlier revisions of it) have been sitting around for a while. If not, no worries.

sameer:
Sure thing! I'll take a look this evening

fitzgen:
Thanks! I really appreciate it.

z3/src/goal.rs
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    /// Return a vector of the formulas from the given goal.
+    pub fn get_formulas<T>(&self) -> Vec<T> where T: Ast<'ctx> {
+        let goal_size = self.get_size();
+        let formula = unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            Z3_goal_formula(self.ctx.z3_ctx, self.z3_goal, 0)
+        };
```
sameer, line 122: Why is the first formula retrieved separately?

z3/src/goal.rs
```diff
@@ -0,0 +1,162 @@
...
+        let mut formulas = vec![T::new(&self.ctx, formula)];
```
sameer, line 123: Since the size is known ahead of time this can be a single allocation:

```rust
Vec::with_capacity(goal_size);
```

z3/src/goal.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// Return the "precision" of the given goal. Goals can be transformed using over and under approximations.
+    pub fn get_precision(&self) -> GoalPrec {
+        let guard = Z3_MUTEX.lock().unwrap();
+        unsafe {
+            Z3_goal_precision(self.ctx.z3_ctx, self.z3_goal)
+        }
+    }
+
+    /// Return a vector of the formulas from the given goal.
+    pub fn get_formulas<T>(&self) -> Vec<T> where T: Ast<'ctx> {
```
sameer, line 117: What about implementing this as an iterator?

Something like this:

```rust
fn iter_formulas(&self) -> impl Iterator<Item = T> {
  (0..goal_size).iter().map(|i| { lock(); Z3_goal_formula(...); })
}
```

It's up to you, just curious what your thoughts on this approach are

z3/src/goal.rs
```diff
@@ -0,0 +1,162 @@
...
+use z3_sys::*;
+
+use crate::ast;
+use crate::ast::Ast;
+use Context;
+use Goal;
+use Z3_MUTEX;
+
+
+impl<'ctx> Goal<'ctx> {
```
sameer, line 13: Is it worth implementing `Clone` for goal with reference counting? It's done for Ast, wondering if it should be done here too

z3/src/tactic.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub fn list_all(ctx: &'ctx Context) {
+        let p = unsafe {
+            Z3_get_num_tactics(ctx.z3_ctx)
+        };
+        for n in 0..p {
+            let t = unsafe {
+                Z3_get_tactic_name(ctx.z3_ctx, n)
+            };
+            match unsafe { CStr::from_ptr(t) }.to_str() {
+                Ok(s) => println!("{}", s),
```
sameer, line 21: did you mean to print this out? maybe we should be returning `Vec<String>` or `impl Iterator<Item=String>` here

z3/src/goal.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub fn new_from_z3_type(ctx: &'ctx Context, z3_goal: Z3_goal) -> Goal<'ctx>{
```
sameer, line 14: does this need to be `pub` to users of the crate? maybe make it `pub(crate)`

z3/src/tactic.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    fn new_from_z3(ctx: &'ctx Context, z3_tactic: Z3_tactic) -> Tactic<'ctx> {
+        Tactic {
+            ctx,
+            z3_tactic,
+        }
+    }
+
+    pub fn new(ctx: &'ctx Context, name: &'ctx str) -> Tactic<'ctx> {
```
sameer, line 34: Can `&'ctx str` be changed to `&str`? Since the str lifetime may not be as short/long as the ctx lifetime, and we're not keeping that reference in the returned tactic struct

z3/src/tactic.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub fn and_then(&self, then_tactic: Tactic) -> Tactic {
+        let t = unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            let t = Z3_tactic_and_then(self.ctx.z3_ctx, self.z3_tactic, then_tactic.z3_tactic);
+            Z3_tactic_inc_ref(self.ctx.z3_ctx, t);
+            t
+        };
+        Tactic {
+            ctx: self.ctx,
+            z3_tactic: t,
```
sameer, line 92: any reason not to keep the unsafe block here?

z3/src/tactic.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub fn or_else(&self, else_tactic: Tactic) -> Tactic {
+        let t = unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            let t = Z3_tactic_or_else(self.ctx.z3_ctx, self.z3_tactic, else_tactic.z3_tactic);
+            Z3_tactic_inc_ref(self.ctx.z3_ctx, t);
+            t
+        };
+        Tactic {
+            ctx: self.ctx,
+            z3_tactic: t,
```
sameer, line 107: same comment here

z3/src/tactic.rs
```diff
@@ -0,0 +1,159 @@
...
+        let result = unsafe {
+            Z3_tactic_apply(
+                self.ctx.z3_ctx,
+                self.z3_tactic,
+                goal.z3_goal,
+            )
+        };
+        unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            Z3_apply_result_inc_ref(self.ctx.z3_ctx, result);
```
sameer, line 121: It looks like it'll be hard to keep track of the reference count for the apply result. You might just have to create an apply result struct that lets you iterate over the subgoals via `impl Iterator<Item = Goal<'ctx>>` and decrements this when dropped.

z3/src/tactic.rs
```diff
@@ -0,0 +1,159 @@
...
+        };
+        Tactic {
+            ctx: self.ctx,
+            z3_tactic: t,
+        }
+    }
+
+    pub fn apply(self, goal: &Goal, params: &Params) -> Goal<'ctx> {
+        let result = unsafe {
+            Z3_tactic_apply(
```
sameer, line 113: I don't see `params` being used here. If params is optional, could you make the signature `params: Option<&Params>`, match on that, and apply the arms `None => Z3_tactic_apply, Some(params) => Z3_tactic_apply_ex`?

z3/tests/lib.rs
```diff
@@ -865,6 +993,136 @@ fn test_array_store_select() {
 
 #[test]
 #[should_panic]
+fn test_goal_get_formulas_empty() {
```
sameer, line 996: is there any particular reason not to return an empty vec when the size is 0 instead of panicking?

z3/src/tactic.rs
```diff
@@ -0,0 +1,159 @@
...
+                self.ctx.z3_ctx,
+                self.z3_tactic,
+                goal.z3_goal,
+            )
+        };
+        unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            Z3_apply_result_inc_ref(self.ctx.z3_ctx, result);
+            if Z3_apply_result_get_num_subgoals(self.ctx.z3_ctx, result) == 1 {
```
sameer, line 122: This will panic when the apply result contains more than 1 (or even 0) subgoals. Can we consider all cases here?

z3/src/tactic.rs
```diff
@@ -0,0 +1,159 @@
...
+use std::fmt;
+
+use z3_sys::*;
+use Context;
+use Goal;
+use Params;
+use Tactic;
+use Z3_MUTEX;
+
+impl<'ctx> Tactic<'ctx> {
```
sameer, line 11: implement Clone for Tactic with ref counting

sameer (changes requested):
Thanks for implementing this. I've requested a few changes inline, if you feel strongly about some feel free to discuss.

Some high level comments:

* Several of the docstrings are verbatim from the z3_sys crate. I'm fine with that but could you make sure the variable references make sense in context of the higher level method signature?
* Few style nitpicks. Feel free to gloss over those if you don't have time.
* I referenced the [Python equivalent](https://github.com/Z3Prover/z3/blob/49aebdbb02ae63e2cf558f6c3d8a40bfed359f9d/src/api/python/z3/z3.py#L5199) to get better context. Let me know if there's another resource you prefer & referred to for this (other than z3_sys). Would be good to have that handy :)

z3/src/goal.rs
```diff
@@ -0,0 +1,162 @@
...
+            z3_goal: z3_goal,
+        }
+    }
+
+    pub fn new(ctx: &'ctx Context, models: bool, unsat_cores: bool, proofs: bool) -> Goal<'ctx> {
+        let goal = Self {
+            ctx,
+            z3_goal: unsafe {
+                let guard = Z3_MUTEX.lock().unwrap();
+                let g = Z3_mk_goal(ctx.z3_ctx, models, unsat_cores, proofs);
```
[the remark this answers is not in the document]
rlkelly: I went through the C++ api to see how it's done: 
https://github.com/Z3Prover/z3/blob/518296dbc10267d4a4b8589212feaeefca800022/src/api/api_goal.cpp#L32
https://github.com/Z3Prover/z3/blob/518296dbc10267d4a4b8589212feaeefca800022/src/api/api_context.h#L145

I'm not sure how to access the `api::context` from `Z3_context`.  If you see a way please let me know.

I also don't see this being done in the python API, so maybe this check can be skipped:
https://github.com/Z3Prover/z3/blob/master/src/api/python/z3/z3.py#L5207

z3/src/goal.rs
rlkelly: That definitely seems like a better option though, since it can be lazily evaluated.  I'm running into some issues with the lifetime but will attempt to implement it.

z3/src/tactic.rs
rlkelly: I left that out since it wasn't strictly necessary.  I'm happy to oblige if that's the preference.

z3/src/tactic.rs
rlkelly: that makes much more sense, I don't recall what my thinking was here!

rlkelly:
@sameer Thanks for your comments, I found them very helpful and insightful.  Let me know how it looks when you get a chance.

z3/src/goal.rs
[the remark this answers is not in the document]
sameer: No worries! Just worth mentioning it in the docs for that method then

z3/src/tactic.rs
```diff
@@ -0,0 +1,191 @@
...
+#[derive(Clone, Debug)]
+pub struct ApplyResult<'ctx> {
+    ctx: &'ctx Context,
+    z3_apply_result: Z3_apply_result,
+}
+
+impl<'ctx> ApplyResult<'ctx> {
+    pub fn list_subgoals(&self) -> impl Iterator<Item = Goal> {
+        unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
```
sameer, line 22: this guard should be the first line in the map closure body, otherwise it would be dropped after this method returns

sameer (changes requested):
Looking pretty good! Few minor changes and it's good to go :slightly_smiling_face:

z3/src/goal.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    /// Return the "precision" of the given goal. Goals can be transformed using over and under approximations.
+    pub fn get_precision(&self) -> GoalPrec {
+        let guard = Z3_MUTEX.lock().unwrap();
+        unsafe {
+            Z3_goal_precision(self.ctx.z3_ctx, self.z3_goal)
+        }
+    }
+
+    fn iter_formulas<'a, T>(&'a self) -> impl Iterator<Item = T> + 'a where T: Ast<'a> {
```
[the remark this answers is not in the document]
rlkelly: I believe so, but I don't want to claim to fully understand the inner workings of the reference counter ;-)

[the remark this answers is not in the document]
rlkelly: it will create a `Z3_ast` struct at each step, so I don't believe it has any impact on the goal itself.

sameer (approved):
Thanks for making those changes, this looks good to me!

fitzgen:
Thanks @rlkelly and thanks @sameer!

Issue #127, opened 2021-02-05
Implement Probe


rlkelly:
Probe is a class in Z3 used to inspect goals and collect information.  This would be a very useful addition.

rlkelly:
I can start on this.  Will make a PR once #124 gets merged

PR #128, opened 2021-02-09
add probe

rlkelly:
This PR is for the probe implementation mentioned in #127 .

### DIFF
--- a/z3/src/probe.rs
+++ b/z3/src/probe.rs
@@ -1,0 +1,218 @@
+use std::ffi::{CStr, CString};
+use std::fmt;
+use std::result::Result;
+use std::str::Utf8Error;
+
+use z3_sys::*;
+use Context;
+use Goal;
+use Probe;
+use Z3_MUTEX;
+
+impl<'ctx> Probe<'ctx> {
+    pub fn list_all(ctx: &'ctx Context) -> impl Iterator<Item=std::result::Result<&'ctx str, Utf8Error>> {
+        let p = unsafe {
+            Z3_get_num_probes(ctx.z3_ctx)
+        };
+        (0..p).into_iter().map(move |n| {
+            let t = unsafe {
+                Z3_get_probe_name(ctx.z3_ctx, n)
+            };
+            unsafe { CStr::from_ptr(t) }.to_str()
+        })
+    }
+
+    pub fn describe(ctx: &'ctx Context, name: &str) -> std::result::Result<&'ctx str, Utf8Error> {
+        let probe_name = CString::new(name).unwrap();
+        unsafe {
+            CStr::from_ptr(Z3_probe_get_descr(ctx.z3_ctx, probe_name.as_ptr())).to_str()
+        }
+    }
+
+    pub fn new(c: &'ctx Context, name: &str) -> Probe<'ctx> {
+        let probe_name = CString::new(name).unwrap();
+        Probe {
+            ctx: c,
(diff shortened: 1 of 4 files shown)


rlkelly:
@fitzgen

z3/src/probe.rs
```diff
@@ -0,0 +1,218 @@
...
+            Z3_probe_inc_ref(self.ctx.z3_ctx, z3_probe);
+            let z3_probe_not = Z3_probe_not(self.ctx.z3_ctx, z3_probe);
+            Z3_probe_inc_ref(self.ctx.z3_ctx, z3_probe_not);
+            Probe {
+                ctx: self.ctx,
+                z3_probe: z3_probe_not,
+            }
+        }
+    }
+}
```
sameer, line 184: These methods really lend themselves to easy [std::ops](https://doc.rust-lang.org/std/ops/) operator implementation over Probe. Would you be able to add that if you have time?

I'm imagining a scenario like
```rust
let probe_a;
let probe_b;
// Nice ergonomics!
probe_a < probe_b
```

sameer (changes requested):
Few minor changes, LGTM overall. Thanks for adding this.

(We are watching this repository so there is no need to mention us, for the record.)

z3/src/probe.rs
```diff
@@ -0,0 +1,218 @@
...
+            Probe {
+                ctx: self.ctx,
+                z3_probe,
+            }
+        }
+    }
+
+    /// Return a probe that evaluates to "true" when the value returned by `self` is not equal to the value returned by `p`.
+    pub fn ne(&self, p: &Probe) -> Probe<'ctx> {
+        unsafe {
```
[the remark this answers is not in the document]
rlkelly: no, this was just something to make things a bit more convenient and mirror the python api.

z3/src/probe.rs
```diff
@@ -0,0 +1,218 @@
...
+                Z3_probe_inc_ref(self.ctx.z3_ctx, self.z3_probe);
+                self.z3_probe
+            },
+        }
+    }
+}
+
+impl<'ctx> fmt::Display for Probe<'ctx> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        write!(f, "<z3.probe>")
```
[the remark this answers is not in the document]
rlkelly: pretty much.  I spent a little time exploring the idea, but it doesn't seem worth the effort

z3/src/probe.rs
rlkelly: yeah I was actually considering that.  I've never used `std::ops` so this is a good time to learn!  I'll try to add that in tomorrow.

rlkelly: at first glance though, it looks like something like this would work:

```
impl<'ctx> Not for Probe<'ctx> {
    type Output = Probe<'ctx>;

    fn not(self) -> Self::Output {
        self.not()
    }
}
```

Since you can overload std::ops, but `std::cmp::PartialEq` seems to require a return type of `bool` : https://doc.rust-lang.org/src/core/cmp.rs.html#203

jagill: Maybe adding `std::ops` support could be another PR?  I think those can be a bit more contentious, while exposing core Z3 functionality seems like a straight win.

sameer: sounds good to me

sameer (approved):

Issue #129, opened 2021-03-04
ast.is_null() error when comparing two dynamic variables


rlkelly:
I'm getting an error `assertion failed: !ast.is_null()`  when running this code:

```
let cfg = Config::new();
let ctx = &Context::new(&cfg);
let x: ast::Dynamic = ast::Bool::new_const(ctx, "a").into();
let y: ast::Dynamic = ast::String::from_str(ctx, "a").unwrap().into();
x._eq(&y);
```

How should this be handled?

sameer:
What are you trying to do with that example code? Is the question about how to handle comparison between dynamic nodes that don't have the same sort?

On Wed, Mar 3, 2021, at 20:26, Rob Kelly wrote:
> 

> I'm getting an error `assertion failed: !ast.is_null()` when running this code:

> `let cfg = Config::new();
let ctx = &Context::new(&cfg);
let x: ast::Dynamic = ast::Bool::new_const(ctx, "a").into();
let y: ast::Dynamic = ast::String::new_const(ctx, "a").into();
x._eq(&y);
`

rlkelly:
I'm curious if you think this could potentially return a `Result` or at least a more verbose error message, or if that should be handled by application logic?

sameer:
> panicking with an error message indicating their types would be much easier to interpret

I think that's a good plan. Assertions made in z3_rs should have error messages with some hints as to what went wrong, and the conditions in which they occur should be called out in docs.

In this case it's because `Ast::_eq` does not return a result so there's a panic for the sake of interface parity. I think that's a fair design decision since the user knows that these are dynamic and they may not have the same sort. But that possibility should be made clear.

PR #130, opened 2021-03-05
add safe_eq

rlkelly:
this is adds a method "_safe_eq" that returns a `Result` when comparing two ast nodes and checks that they have the same `SortKind`.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -6,6 +6,7 @@ use std::hash::{Hash, Hasher};
 use z3_sys::*;
 use Context;
 use FuncDecl;
+use IsNotApp;
 use Pattern;
 use Sort;
 use SortDiffers;
@@ -336,16 +337,21 @@ pub trait Ast<'ctx>: Sized + fmt::Debug + Into<Dynamic<'ctx>> {
     /// This will panic if the `Ast` is not an app, i.e. if AstKind is not App
     /// or Numeral.
     fn decl(&self) -> FuncDecl<'ctx> {
-        let ctx = self.get_ctx();
-        let func_decl = unsafe {
-            let guard = Z3_MUTEX.lock().unwrap();
-            let app = Z3_to_app(ctx.z3_ctx, self.get_z3_ast());
-            assert!(!app.is_null(), "Ast has a null Z3_app.");
-            let func_decl = Z3_get_app_decl(ctx.z3_ctx, app);
-            assert!(!func_decl.is_null(), "Ast has a null Z3_func_decl.");
-            func_decl
-        };
-        unsafe { FuncDecl::from_raw(ctx, func_decl) }
+        self.safe_decl().expect("Ast is not an app")
+    }
+
+    fn safe_decl(&self) -> Result<FuncDecl<'ctx>, IsNotApp> {
+        if !self.is_app() {
+            Err(IsNotApp::new(self.kind()))
+        } else {
+            let ctx = self.get_ctx();
+            let func_decl = unsafe {
+                let guard = Z3_MUTEX.lock().unwrap();
+                let app = Z3_to_app(ctx.z3_ctx, self.get_z3_ast());
+                Z3_get_app_decl(ctx.z3_ctx, app)
+            };
+            Ok(unsafe { FuncDecl::from_raw(ctx, func_decl) })
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


z3/src/ast.rs
```diff
@@ -196,6 +196,25 @@ pub trait Ast<'ctx>: Sized + fmt::Debug + Into<Dynamic<'ctx>> {
...
+        let left_sort = unsafe {
+            Z3_get_sort_kind(ctx, Z3_get_sort(ctx, self.get_z3_ast()))
+        };
+        let right_sort = unsafe {
+            Z3_get_sort_kind(other_ctx, Z3_get_sort(other_ctx, other.get_z3_ast()))
+        };
+        if (left_sort == right_sort) {
+            Ok(self._eq(other))
+        } else {
+            Err(format!("Can not compare nodes, Sort does not match.  Nodes contain types {:?} and {:?}", left_sort, right_sort))
```
sameer, line 214: I agree that it's implied that the two don't have the same sort but I don't think the Err type should be a string. Maybe better suited as a struct whose display trait impl is the error message. Like

```rust
struct SortDiffers<'ctx> {
  left: Sort<'ctx>, right: Sort<'ctx>
}

impl<'ctx> Display for Sort<'ctx> { ... }
```

z3/src/ast.rs
```diff
@@ -196,6 +196,25 @@ pub trait Ast<'ctx>: Sized + fmt::Debug + Into<Dynamic<'ctx>> {
         })
     }
 
+
+    /// Compare this `Ast` with another `Ast`, and get a Result.  Errors if the sort does not
+    /// match for the two values.
+    fn _safe_eq(&self, other: &Self) -> Result<Bool<'ctx>, std::string::String> {
```
sameer, line 202: now that there's a safe method could you make `_eq` call it? That way z3 can unwrap with the error message instead of just the opaque assertion.

sameer (changes requested):

z3/src/lib.rs
```diff
@@ -17,7 +17,7 @@ extern crate num;
 use std::ffi::CString;
 use std::sync::Mutex;
 use z3_sys::*;
-pub use z3_sys::GoalPrec;
+pub use z3_sys::{AstKind, GoalPrec, SortKind};
```
rlkelly, line 20: I just want to point this out, I hope this is alright.

rlkelly:
#129

z3/src/ast.rs
```diff
@@ -319,16 +337,23 @@ pub trait Ast<'ctx>: Sized + fmt::Debug + Into<Dynamic<'ctx>> {
...
+    fn safe_decl(&self) -> Result<FuncDecl<'ctx>, IsNotApp> {
+        if !self.is_app() {
+            Err(IsNotApp::new(self.kind()))
+        } else {
+            let ctx = self.get_ctx();
+            let func_decl = unsafe {
+                let guard = Z3_MUTEX.lock().unwrap();
+                let app = Z3_to_app(ctx.z3_ctx, self.get_z3_ast());
+                let func_decl = Z3_get_app_decl(ctx.z3_ctx, app);
+                assert!(!func_decl.is_null(), "Ast has a null Z3_func_decl.");
```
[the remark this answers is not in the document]
rlkelly: it was a sanity check, we can get rid of it though

rlkelly:
I added three small changes.  I removed clone from tactic since this will raise reference count errors, made tactics take references, and made `list_subgoals` move the `ApplyResult`.

sameer:
Will review once I'm done with work but the changes sound reasonable 🙂

z3/src/tactic.rs
```diff
@@ -4,43 +4,25 @@ use std::result::Result;
...
-                let sg = Z3_apply_result_get_subgoal(self.ctx.z3_ctx, self.z3_apply_result, i);
-                Z3_goal_inc_ref(self.ctx.z3_ctx, sg);
-                Goal::new_from_z3_type(self.ctx, sg, true, true, true)
-            })
-        }
-    }
-}
-
-impl<'ctx> Clone for Tactic<'ctx> {
-    fn clone(&self) -> Self {
```
sameer, line 35: I understand the point of the change and I'm fine with it, but in the future it would be better-suited to a separate PR

sameer (approved):
thanks for powering through these changes, looks great! :rocket:

PR #135, opened 2021-03-24
z3-sys: automate enum generation via bindgen

sameer:
The benefit here is adhering to the interface with older versions of z3. If the Rust bindings don't support it, it'll fail to build instead of
transiently failing at runtime.


sameer:
v4.8.10 broke submodules that aren't in the root directory of a git repo so I couldn't include that in this PR unfortunately

PR #136, opened 2021-03-25
z3: add docstrings for all top-level items

lucab:
This updates crate documentation by adding docstring entries to all
top-level API items.

### DIFF
--- a/z3-sys/src/lib.rs
+++ b/z3-sys/src/lib.rs
@@ -350,7 +350,7 @@ pub enum ParameterKind {
     FuncDecl = generated::Z3_parameter_kind::Z3_PARAMETER_FUNC_DECL as u32,
 }
 
-/// The different kinds of Z3 types (See [`Z3_get_sort_kind`](fn.Z3_get_sort_kind.html)).
+/// The different kinds of Z3 types.
 ///
 /// This corresponds to `Z3_sort_kind` in the C API.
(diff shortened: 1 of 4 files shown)


z3-sys/src/lib.rs
```diff
@@ -350,7 +350,7 @@ pub enum ParameterKind {
     FuncDecl = generated::Z3_parameter_kind::Z3_PARAMETER_FUNC_DECL as u32,
 }
 
-/// The different kinds of Z3 types (See [`Z3_get_sort_kind`](fn.Z3_get_sort_kind.html)).
+/// The different kinds of Z3 types.
 ///
 /// This corresponds to `Z3_sort_kind` in the C API.
```
waywardmonkeys, line 355: Can this link to the docs for `Z3_sort_kind` in `z3-sys`? (Are there docs?)

lucab: To the best of my knowledge, not directly. Indirectly it can point to an absolute URL for a given z3-sys version, but then it becomes tedious to update.

This change is only meant to remove the previously wrong/dead link.

waywardmonkeys: Okay. That might be why I didn't do it orginally then. (I think I did that original "This corresponds to ..." stuff.)

waywardmonkeys:
Thanks!

Issue #142, opened 2021-04-19
About Z3_MUTEX


black-binary:
I'm currently developing a multi-threaded program, but it seems like almost every function call to the z3 library is mutex-guarded. My threads will frequently create expressions and eval() them, and each thread has its own Context (unsafe impl Send for Context)

I'm wondering if the Z3_MUTEX will slow down my code. If so, is it safe to remove the Z3_MUTEX in my code? Thanks :)

waywardmonkeys:
The last time that I looked at this, a couple of years ago now, I guess, it was necessary because of bugs in Z3, perhaps on macOS. I think those bugs were fixed, but I don't know if there are new ones or not.

You might be able to investigate whether or not it can be removed. :)

black-binary:
@waywardmonkeys Thanks for your reply. I plan to run a stress test parallelly after removing the mutex, and see if it will crash.

waywardmonkeys:
This has been removed now in #211 (follow on to #143).

PR #143, opened 2021-04-27
Remove Z3_MUTEX

black-binary:
#142

I removed the global mutex and wrote a simple stress test for it. The stress test spawns 10 threads and runs all the other unit tests for 60 seconds. There is no sign of crash or test failure in my machine(x86, Arch Linux / Ubuntu 20.04). And I think it's ok to remove the mutex now. :)

### DIFF
--- a/z3/Cargo.toml
+++ b/z3/Cargo.toml
@@ -23,7 +23,6 @@ static-link-z3 = ["z3-sys/static-link-z3"]
 
 [dependencies]
 log = "0.4"
-lazy_static = "1"
 
 # optional dependencies
 num = { version = "0.2.0", optional=true }
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -13,7 +13,6 @@ use Pattern;
 use Sort;
 use SortDiffers;
 use Symbol;
-use Z3_MUTEX;
 
 pub use z3_sys::AstKind;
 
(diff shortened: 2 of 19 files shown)


waywardmonkeys:
Pretty sure this was an issue on macOS. Perhaps related to the usage of OpenMP which wasn't enabled on macOS. OpenMP is gone from Z3 now, but this will need testing on more than just Linux.

black-binary:
I have no macOS devices so I cannot test that personally. However, the stress test is included in this commit. You can run `cargo test` and check if it works in macOS.

GreenBeard:
I am looking forward to this set of changes being merged. I have a macOS VM that I could run the tests on if that would be acceptable for testing purposes.

jder:
I just cloned this branch and ran `cargo check` on my mac laptop. Everything passed including the 3 parallel stress tests, and it was using ~8-9 cores during those periods.

sameer:
The only lingering concern I have before merging this is the reference counters -- do we know that they are atomic? 

Other than that, I can't think of anything major that would require keeping the lock.

waywardmonkeys:
@sameer The reference counters in most of Z3 shouldn't matter if they're atomic or not since we shouldn't be supporting arbitrarily using most things from any thread.

I think the issue that was solved upstream had to do with other stuff inside of Z3, so we should be able to resurrect this patch and get the removal of Z3_MUTEX landed.

waywardmonkeys:
I'll open a new PR for a new version of this patch.

PR #145, opened 2021-05-09
Removed Dynamic parameters, and refactored some code

GreenBeard:
I have removed `Dynamic` *parameters* (there are still `Dynamic` return types although I did consider changing them into a `Box<&dyn Ast>`) which added a great deal of unnecessary `into()`, and `from()`s in the user's code and Rust has better methods for dealing with this (mainly `&dyn` traits). Therefore I have modified `Ast` to be usable in a `&dyn` context by removing its `Size` trait which also required removing its `Into<Dynamic<'ctx>>`. This isn't an issue as I we have `from_ast` available in the `Dynamic` struct, and the macro `impl_from_try_into_dynamic`. I don't believe that it is possible to directly have a `From` trait that takes in `&dyn Ast` from my understanding, and testing of Rust although I think we have sufficiently covered the needed use cases anyway.

Minor points
- Fixed comment from double to triple slash
- Fixed tests to not use the `into`s
- Fixed `translate` to be on the Ast trait
- Changed `forall_const` to take, and return a `Bool` I searched extensively (in z3, and smtlib2 documentation) to find when it would return something other than `Bool` and I was unable to find information, and I didn't like the unneeded `into`, but if it turns out that it can return something other than `Bool` I will happily change it.
- Fixed `eval` to accept a `model_completion` parameter as some code I am writing needed it (and the C version has this parameter). I probably should have done this in a different pull request, but it seemed minor enough to include here with the other clean ups.
- Fixed some warnings
- Enabled test `test_set_membership`
- Added `where Self:Sized` when Rust complained on certain `Ast` trait functions. What it effectively means is that it requires a specific instance of `Ast` as opposed to any generic one. For example: this can require multiple elements in a slice to be of the same type.

I have a few quick questions:
- Does `simplify` always return the same type as the input? For example can a `Z3_func_decl` returning 1 in z3 be simplified to a `Z3_int` which would change its type?
- Can/should `substitute` allow multiple different types as currently it requires that each substitution pair be the same type. An example that currently isn't allowed is `(0, 1), ('a', 'b')` in psuedocode as they have different types, but each by themselves would be fine.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -162,13 +162,13 @@ macro_rules! varop {
 
 /// Abstract syntax tree (AST) nodes represent terms, constants, or expressions.
 /// The `Ast` trait contains methods common to all AST subtypes.
-pub trait Ast<'ctx>: Sized + fmt::Debug + Into<Dynamic<'ctx>> {
+pub trait Ast<'ctx>: fmt::Debug {
     fn get_ctx(&self) -> &'ctx Context;
     fn get_z3_ast(&self) -> Z3_ast;
 
     // This would be great, but gives error E0071 "expected struct, variant or union type, found Self"
     // so I don't think we can write a generic constructor like this.
-    // Instead we just require the method, and use the new_ast! macro defined below to implement it
+    // Instead we just require the method, and use the impl_ast! macro defined below to implement it
     // on each Ast subtype.
     /*
     fn new(ctx: &'ctx Context, ast: Z3_ast) -> Self {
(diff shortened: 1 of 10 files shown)


GreenBeard:
Hmm, it looks like there were more tests than I realized. (I had only used `cargo test`.) I will go back, and fix these today, or tomorrow. I assume that most of the issues just require changing the usage of `Dynamic`, and other minor tweaks of some code, but I won't know until investigating.

sameer:
I unfortunately don't have the answers to your questions, but the PR overall looks good. If I understood correctly, there are breaking changes here so I will do a minor version bump for the next release.

GreenBeard:
@sameer Yep, the changes are breaking. Maybe I'll look into asking the z3 maintainers about my questions. However, I was distracted with other things for a while which happen again :)

sameer:
No worries! Thanks again for making these changes :rocket:

PR #148, opened 2021-06-25
Add interpreted (recursive) functions

skius:
This PR adds wrapping to the `z3` crate of `z3-sys`'s `Z3_mk_rec_func_decl` and `Z3_add_rec_def`.

I implemented a new `RecFuncDecl` struct which has an `add_def(args, body)` method, since the `z3-sys` API differentiates between `mk_func_decl` and `mk_rec_func_decl` as far as I can tell.

The `add_def` method has a doc example on how to use the new API and there are two tests (`test_rec_func_def` and `test_rec_func_def_unsat`) in `z3/test/lib.rs`.

I'm not familiar at all with unsafe Rust code and only have a very basic understanding of SMT solvers, so please take a good look and let me know if something's not up to standard.

### DIFF
--- a/z3/src/rec_func_decl.rs
+++ b/z3/src/rec_func_decl.rs
@@ -79,8 +79,10 @@ impl<'ctx> RecFuncDecl<'ctx> {
         args: &[&ast::Dynamic<'ctx>],
         body: &impl Ast<'ctx>,
     ) {
-        let mut args: Vec<_> = args.iter().map(|s| s.get_z3_ast()).collect();
+        assert!(args.iter().all(|arg| arg.ctx == body.get_ctx()));
+        assert_eq!(self.ctx, body.get_ctx());
 
+        let mut args: Vec<_> = args.iter().map(|s| s.get_z3_ast()).collect();
         unsafe {
             assert_eq!(body.get_sort().z3_sort, Z3_get_range(self.ctx.z3_ctx, self.z3_func_decl));
 
--- a/z3/src/rec_func_decl.rs
+++ b/z3/src/rec_func_decl.rs
@@ -1,0 +1,184 @@
+use ast;
+use ast::Ast;
+use std::convert::TryInto;
+use std::ffi::CStr;
+use std::fmt;
+use z3_sys::*;
+use {Context, RecFuncDecl, Sort, Symbol, Z3_MUTEX};
+
+impl<'ctx> RecFuncDecl<'ctx> {
+    pub fn new<S: Into<Symbol>>(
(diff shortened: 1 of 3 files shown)


z3/src/rec_func_decl.rs
```diff
@@ -0,0 +1,184 @@
...
+    ///
+    /// solver.assert(&forall);
+    /// let res = solver.check();
+    /// assert_eq!(res, SatResult::Sat);
+    /// ```
+    ///
+    /// Note that `args` should have the types corresponding to the `domain` of the `RecFuncDecl`.
+    pub fn add_def(
+        &self,
+        args: &[&ast::Dynamic<'ctx>],
```
sameer, line 79: do we need to assert on args & body being in the same ctx?

z3/src/rec_func_decl.rs
```diff
@@ -0,0 +1,184 @@
...
+impl<'ctx> fmt::Debug for RecFuncDecl<'ctx> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        <Self as fmt::Display>::fmt(self, f)
+    }
+}
+
+impl<'ctx> Drop for RecFuncDecl<'ctx> {
+    fn drop(&mut self) {
+        unsafe {
+            Z3_dec_ref(
```
sameer, line 178: This might need a mutex lock -- not sure that z3 reference counters are atomic.

sameer (changes requested):
Hi Niels,

Thanks for the PR, this looks really good. Just 2 minor comments and we can merge this.

Best,
Sameer

z3/src/rec_func_decl.rs
skius: I'm not sure either, I copied that impl from FuncDecl https://github.com/prove-rs/z3.rs/blob/8b77fc554def507162a3860bea659247f2939859/z3/src/func_decl.rs#L128-L134 - should I aquire Z3_MUTEX?

z3/src/rec_func_decl.rs
skius: I can't comment on this, I don't really know where/how one would use multiple contexts.

z3/src/rec_func_decl.rs
sameer: Ah ok, that's fine then. We can continue that thread in #143

sameer (approved):
LGTM! Will merge after the workflow approves

skius:
Great, thank you!

Issue #154, opened 2021-08-16
Support for "try_for" tactic


tyleretzel:
This would allow setting a time limit on tactic execution. It should be a pretty simple wrapper around [this](https://docs.rs/z3-sys/0.7.0/z3_sys/fn.Z3_tactic_try_for.html) function in z3-sys: I'll put up a pull request as well!

sameer:
Resolved by #160

PR #155, opened 2021-08-16
Binding for `try_for` tactic

tyleretzel:
https://github.com/prove-rs/z3.rs/issues/154

### DIFF
--- a/z3/src/tactic.rs
+++ b/z3/src/tactic.rs
@@ -2,6 +2,9 @@ use std::ffi::{CStr, CString};
 use std::fmt;
 use std::result::Result;
 use std::str::Utf8Error;
+use std::time::Duration;
+use std::os::raw::c_uint;
+use std::convert::TryFrom;
 
 use z3_sys::*;
 use ApplyResult;
@@ -100,6 +103,21 @@ impl<'ctx> Tactic<'ctx> {
                 Z3_tactic_inc_ref(ctx.z3_ctx, t);
                 t
             },
+        }
+    }
+
+    /// Return a tactic that applies the current tactic to the goal, failing if
+    /// `t` doesn't terminate within the period specified by `timeout`.
+    pub fn try_for(&self, timeout: Duration) -> Tactic<'ctx> {
+        unsafe {
+            let guard = Z3_MUTEX.lock().unwrap();
+            let timeout_ms = c_uint::try_from(timeout.as_millis()).unwrap_or(c_uint::MAX);
+            let t = Z3_tactic_try_for(self.ctx.z3_ctx, self.z3_tactic, timeout_ms);
+            Z3_tactic_inc_ref(self.ctx.z3_ctx, t);
+            Tactic {
+                ctx: self.ctx,
+                z3_tactic: t,
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


z3/tests/lib.rs
```diff
@@ -1128,6 +1128,30 @@ fn test_tactic_skip() {
...
+    let two = ast::Int::from_i64(&ctx, 2);
+    let x = ast::Int::new_const(&ctx, "x");
+
+    let goal = Goal::new(&ctx, false, false, false);
+    goal.assert(&x.ge(&one.add(&two)));
+
+    let tactic = Tactic::new(&ctx, "simplify");
+    let try_for_tactic = tactic.try_for(4294967295);
+
+    // Test that the `try_for` can successfully apply underlying tactic
```
tyleretzel, line 1147: It'd be great to have a tactic that reliably uses more than `n` milliseconds to execute (then we could test the failure case here), but I couldn't find anything that fits this bill

z3/src/tactic.rs
```diff
@@ -103,6 +103,20 @@ impl<'ctx> Tactic<'ctx> {
         }
     }
 
+    /// Return a tactic that applies the current tactic to the goal, failing if
+    /// `t` doesn't terminate in `timeout_ms` milliseconds.
+    pub fn try_for(&self, timeout_ms: u32) -> Tactic<'ctx> {
```
sameer, line 108: Could we use [Duration](https://doc.rust-lang.org/std/time/struct.Duration.html) here?

z3/tests/lib.rs
```diff
@@ -1128,6 +1128,30 @@ fn test_tactic_skip() {
...
+
+    let one = ast::Int::from_i64(&ctx, 1);
+    let two = ast::Int::from_i64(&ctx, 2);
+    let x = ast::Int::new_const(&ctx, "x");
+
+    let goal = Goal::new(&ctx, false, false, false);
+    goal.assert(&x.ge(&one.add(&two)));
+
+    let tactic = Tactic::new(&ctx, "simplify");
+    let try_for_tactic = tactic.try_for(4294967295);
```
sameer, line 1145: Please use the associated const [`u32::MAX`](https://doc.rust-lang.org/std/primitive.u32.html#associatedconstant.MAX) for this

z3/tests/lib.rs
sameer: I'm not aware of a tactic like that, but testing success should be good enough

sameer (changes requested):
Thanks for submitting this! Just had a few comments -- after that it should be good to go

tyleretzel:
I did some more testing and tried to actually use this, and I think there's some more fundamental issues with the tactics API: it doesn't seem like we handle tactic "failure" correctly (it always causes a segfault). Even something simple like this will trigger it:

```
#[test]
fn test_tactic_fail() {
    let cfg = Config::new();
    let ctx = Context::new(&cfg);
    let params = Params::new(&ctx);

    let a = ast::Bool::new_const(&ctx, "a");
    let goal = Goal::new(&ctx, false, false, false);
    goal.assert(&a);

    let goal_results = apply_results.list_subgoals().collect::<Vec<Goal>>();
    let goal_result = goal_results.first().unwrap();
    assert_eq!(goal_result.get_formulas::<Bool>(), vec![a],);
}
```

Let me try to see what's going on here and fix this, too.

sameer:
Good find, keep me posted! :slightly_smiling_face:

PR #160, opened 2021-09-12
Support tactic failure and `try_for` tactic

tyleretzel:
This includes #155, and also handles the nullptr case from `Z3_tactic_apply` when tactics fail.


sameer (approved):
LGTM, thanks Tyler!

PR #161, opened 2021-09-16
Add option to choose z3.h location for bindgen

futile:
This enables the user to choose the path to their z3.h, and doesn't force usage of the system header through wrapper.h.

I.e., by setting the environment variable `Z3_SYS_Z3_HEADER`, the user can determine which z3.h will be used.

Needed this in a project where we did not want to just use the system z3.

### DIFF
--- a/z3-sys/README.md
+++ b/z3-sys/README.md
@@ -24,6 +24,19 @@ Add it to your `Cargo.toml` like so:
 z3-sys = "0.7.1"
 ```
 
+**Note:** This crate requires a `z3.h` during build time.
+
+* By default, the crate will look for a `z3.h` in standard/system include paths.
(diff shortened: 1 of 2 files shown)


sameer (changes requested):
The use case makes sense to me, thanks for making this change! :rocket:

One minor thing: could we document it in the README for z3-sys? That way others can find the option.

futile:
> The use case makes sense to me, thanks for making this change! 
> 
> One minor thing: could we document it in the README for z3-sys? That way others can find the option.

Thanks for the feedback! I added a section to z3-sys/README.md, does that fit?

sameer:
Looks good to me :slightly_smiling_face:

sameer (approved):

PR #168, opened 2021-11-01
Modify z3

rlkelly:


waywardmonkeys:
The relevant parts of this landed some time ago.

PR #176, opened 2021-12-08
Remove #[allow(dead code)] and #[allow(unused variables)]

SeeSpring:
This also fixes an error in _safe_eq

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -93,7 +93,7 @@ macro_rules! unop {
             $( #[ $attr ] )*
             pub fn $f(&self) -> $retty {
                 <$retty>::new(self.ctx, unsafe {
-                    let guard = Z3_MUTEX.lock().unwrap();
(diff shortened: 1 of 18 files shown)


z3/src/ast.rs
```diff
@@ -201,14 +201,14 @@ pub trait Ast<'ctx>: fmt::Debug {
     /// Compare this `Ast` with another `Ast`, and get a Result.  Errors if the sort does not
     /// match for the two values.
     fn _safe_eq(&self, other: &Self) -> Result<Bool<'ctx>, SortDiffers<'ctx>> where Self: Sized {
-        let ctx = self.get_ctx().z3_ctx;
-        let other_ctx = other.get_ctx().z3_ctx;
+        assert!(self.get_ctx() == other.get_ctx());
```
sameer, line 204: minor: any reason not to use assert_eq! here?

sameer (commented):
Looks good overall, thanks for doing this!

PR #177, opened 2021-12-08
Silence warnings from unused enums

SeeSpring:
Some of the generated enum variants are not reexported, this silences the warning

(this hides the unsoundness in z3-sys/lib.rs though)


SeeSpring:
Fixed in #205

PR #178, opened 2021-12-08
Mark Ast::new as unsafe

### LINKED ISSUE #170: Unsafe function not marked as such
[https://github.com/prove-rs/z3.rs/blob/master/z3/src/ast.rs#L187](https://github.com/prove-rs/z3.rs/blob/master/z3/src/ast.rs#L187) should be marked as `unsafe` due to the fact that it uses a pointer that may be invalid. I realize that it checks if the pointer is null, but code such as the following shouldn't be allowed in `safe` Rust as it will invoke undefined behavior when run (in this case it causes a segmentation fault).

```
use z3::Config;
use z3::Context;
use z3::ast::Ast;
use z3_sys::Z3_ast;

fn main() {
  let config = Config::new();
  let context = Context::new(&config);
  z3::ast::Bool::new(&context, 5 as Z3_ast);
}
```

I can fix this if you are would accept a pull request for this. I'm a bit distracted with school as per usual, so I will probably take a week, or two to make a pull request.

SeeSpring:
Fixes #170

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -92,10 +92,12 @@ macro_rules! unop {
         $(
             $( #[ $attr ] )*
             pub fn $f(&self) -> $retty {
-                <$retty>::new(self.ctx, unsafe {
-                    let _guard = Z3_MUTEX.lock().unwrap();
-                    $z3fn(self.ctx.z3_ctx, self.z3_ast)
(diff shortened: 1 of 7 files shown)


waywardmonkeys:
This got done as part of #209.

PR #181, opened 2021-12-09
Removed unneeded bindgen features

SeeSpring:
Some features of bindgen are not needed when building the library and
lead to unused dependencies (mainly `clap`). This change deactivates all
`bindgen` features except for `runtime` which is required for build to
work.

See also https://github.com/rust-rocksdb/rust-rocksdb/pull/491


sameer:
Thanks!

PR #184, opened 2022-01-20
expose Z3_mk_solver_for_logic

dzhu:
This simply adds a small wrapper for Z3_mk_solver_for_logic, similar to
the one for Z3_mk_solver (though it returns an Option<Solver>, since the
input may be invalid).

(Note that Z3_mk_solver_for_logic returns a null pointer on invalid
input, contrary to its comment, which suggests that it returns a default
solver.)

### DIFF
--- a/z3/src/solver.rs
+++ b/z3/src/solver.rs
@@ -8,6 +8,7 @@ use Model;
 use Params;
 use SatResult;
 use Solver;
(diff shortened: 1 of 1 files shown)


sameer (approved):
LGTM! Just going to wait for the workflow to run before merging :rocket:

PR #188, opened 2022-03-09
Correcting operations used for Floating-Point sorts ge and le operations

abhishekc-sharma:
Floating point sorts using incorrect operations for "greater than or equal" and "less than or equal" operations.


sameer:
Good catch, thanks for fixing this!

Issue #190, opened 2022-04-12
Constructing an uninterpreted sort value?


dijkstracula:
My apologies if I am driving z3.rs incorrectly; I'm new to Rust and this library, and reasonably new to SMT solvers in general.  Please let me know if my question is nonsense!

I'd like to write code in z3.rs more or less equivalent to the following smt2 code:

```smt
(declare-sort Node 0)
(declare-fun lock_msg (Node) Bool)

; mutual exclusion: no two clients think they hold the lock simultaneously
(assert (forall ((n1 Node) (n2 Node))
  (=> (and (lock_msg n1) (lock_msg n2)) (= n1 n2)))
```

In Python, I'd probably do it like so:

```python
node = z3.DeclareSort('node')
lock_msg = z3.Function('lock_msg', node, z3.BoolSort())

N1, N2 = z3.Consts('N1 N2', node)
s.assert(z3.Forall([N1, N2], z3.And(v.grant_msg(N1), v.holds_lock(N2))))
```

Defining a new sort and the lock_msg function in Rust-land is reasonably straightforward, but in order to construct the universal I need to be able to supply bounds and a pattern, which seems to necessitate constructing an uninterpreted value as I did in Python.  Thinking that `ast::Datatype::new_const` would do the trick, I wrote something like the following:

```rust
    let node_sort = z3::Sort::uninterpreted(&ctx, z3::Symbol::from("node"));
    let lock_msg = z3::FuncDecl::new(&ctx, "lock_msg", &[&node_sort], &z3::Sort::bool(&ctx));

...
    let n1 = z3::ast::Datatype::new_const(&ctx, "N1", &node_sort);
    let n2 = z3::ast::Datatype::new_const(&ctx, "N2", &node_sort);

    let n1_n2_pattern = z3::Pattern::new(&ctx, &[&n1, &n2]);
...
        &z3::ast::forall_const(&ctx, &[&n1, &n2], &[&n1_n2_pattern], 
            &z3::ast::Bool::and(&ctx, &[&lock_msg.apply(&[&n1]).as_bool().unwrap(),
                                    &lock_msg.apply(&[&n2]).as_bool().unwrap()])
                .implies(&n1._eq(&n2)))
```

Constructing n1 panics, however:

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Uninterpreted`,
[...]

PR #191, opened 2022-04-21
Implement constructing consts of an uninterpreted sort

dijkstracula:
My attempt at addressing https://github.com/prove-rs/z3.rs/issues/190 .

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -83,6 +83,12 @@ pub struct Dynamic<'ctx> {
     pub(crate) z3_ast: Z3_ast,
 }
 
+/// [`Ast`](trait.Ast.html) node representing an uninterpreted value.
+pub struct Uninterpreted<'ctx> {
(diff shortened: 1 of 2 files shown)


HKalbasi:
Is a new ast type needed? I think a `new_const` function for `Dynamic` would do the job.

dijkstracula:
Apologies, I've been ignoring my github notifications over the break.  Do you mean that `as_uninterpreted()` and `new_const()` should return a `Dynamic`, removing the new Rust datatype?  If you prefer that I can make that change.

HKalbasi:
I don't really understand the need for `as_uninterpreted()`, since it doesn't enable anything more than `Dynamic` itself. Do you have a specific use case? For example in case of `as_int()`, you will get a `Int`, which you can add it to other `Int`s or call `Int` specific functions on it, but there isn't such functions on `Uninterpreted`. So I think it is possible to remove `Uninterpreted` and `as_uninterpreted` without losing functionality.

Issue #192, opened 2022-05-10
builds on windows are broken and subpar in general


Trolldemorted:
During a train ride through rural Europe I had some trouble to use z3-sys due to several reasons, so I thought I should open an issue and PR to streamline the overall user experience of developers.

I have identified a few problems:
- Building z3-sys requires a bespoken clang installation. I installed clang via the visual studio installer and it works nice in cmd and powershell, but your current build process does not find it and fails.
- Building z3-sys requires a clang installation at all, despite being a shallow wrapper around the native library. I know having bindgen in your build script has its advantages, but it should not be in the default case, but rather hidden behind a feature.
- The installation documentation is incomplete and partially misleading.
- Using precompiled Z3 binaries is not supported. Using z3-sys on weak/old systems imposes a long initial build time (*sad old laptop noises*).
- Several users were unable to build z3-sys (#183, #87) due to the issues identified above.

I'll open a PR that tries to address/amend some of these.

toolCHAINZ:
These (very legitimate) issues have all been resolved now I think.

The package now supports linking against existing z3 installations, using `vcpkg`, building the z3 source (which I believe was the onerous old default), and building against the hosted build artifact from the Z3 github project (#352).

Closing this issue as completed.

Issue #183, opened 2021-12-22
Can't build z3-sys on windows


hermanventer:
```
Error: failed to run custom build command for `z3-sys v0.7.1`
Caused by:
  process didn't exit successfully: `D:\a\MIRAI\MIRAI\target\release\build\z3-sys-e5c5536e9e865e98\build-script-build` (exit code: 101)
  --- stdout
  cargo:rerun-if-changed=build.rs
  cargo:rerun-if-changed=wrapper.h

  --- stderr
  thread 'main' panicked at 'Unable to find libclang: "the `libclang` shared library at C:\\msys64\\mingw64\\bin\\libclang.dll could not be opened: LoadLibraryExW failed"', 
C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\bindgen-0.58.1\src/lib.rs:2057:31
```

I checked and the file does exist:
```
Run ls -l C:\\msys64\\mingw64\\bin

    Directory: C:\msys64\mingw64\bin

Mode                 LastWriteTime            Length   Name
----                 -------------            ------   ----
.......
-a---                12/17/2021  4:13 AM      30677555 libclang.dll
.......
```

hermanventer:
Tried this again and the error is different now:

Error: failed to compile `mirai v1.1.0 (D:\a\MIRAI\MIRAI\checker)`, intermediate artifacts can be found at `D:\a\MIRAI\MIRAI\target`
Caused by:
  failed to run custom build command for `z3-sys v0.7.1`

Caused by:
  process didn't exit successfully: `D:\a\MIRAI\MIRAI\target\release\build\z3-sys-d2d77d3ff8918670\build-script-build` (exit code: 101)
  --- stdout
  cargo:rerun-if-changed=build.rs
  cargo:rerun-if-changed=wrapper.h

  --- stderr
  wrapper.h:1:10: fatal error: 'z3.h' file not found
  wrapper.h:1:10: fatal error: 'z3.h' file not found, err: true
  thread 'main' panicked at 'Unable to generate bindings: ()', C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\z3-sys-0.7.1\build.rs:33:14

hermanventer:
This seems a bit stale now. Not sure it is fixed, but I've moved on.

Issue #172, opened 2021-11-14
z3-sys cannot find z3.h


DavePearce:
Hey,

So, following the docs it says to resolve this configure `Z3_SYS_Z3_HEADER` to the actual path.  But, that doesn't seem to be resolving the issue.  Here are my commands:

```
> setenv Z3_SYS_Z3_HEADER /home/.../include/z3.h
> echo $Z3_SYS_Z3_HEADER
 /home/.../include/z3.h
> cargo run
```

I tried doing `cargo clean` first, but not difference.  Did I miss something?

DavePearce:
I should add that I'm actually linking against `z3` not `z3-sys`.

jmc-figueira:
It seems to me that the issue is that the option to use that environment variable to specify the path explicitly was added after the latest published version of the library. Hence, you need to use the version from the latest git commit (`ef6ca05cac901795112444fd805f25b06a1d5526`) to be able to use that.

dcsommer:
It would be great to enhance the path finding in general to not require users manually specify the path by environment variables. One place to start for inspiration could be this CMake script:

https://github.com/klee/klee/blob/master/cmake/modules/FindZ3.cmake

Ablu:
Would something like https://github.com/rust-lang/rust-bindgen/issues/1247 maybe be a more flexible approach? That would also fix the auto-detection on Fedora (which has the header under `/usr/include/z3/z3.h`). @futile would a pkg-config based detection have worked in your situation? (mentioning since you opened the PR for the configurable header path)

Ablu:
I digged around a bit. The OpenCV wrapper has something pretty sophisticated. I am not suggesting to copy it 1:1, but taking a peek there was certainly interesting: https://github.com/twistedfall/opencv-rust/blob/master/build/library.rs. They derive stuff from all kind of sources (environment variables, pkg_config, cmake, vcpkg). I think either cmake or pkg_config would be a great start and solve 90%+ of the problems?

futile:
> @futile would a pkg-config based detection have worked in your situation? (mentioning since you opened the PR for the configurable header path)

Hey! :) I don't think a pkg-config based detection would have worked in my situation, as I built z3 from source as well (possibly in different configurations), and thus just wanted it to point at the correct file. So I think it would be nice to keep the environment override around in some form, but pkg-config and/or cmake detection capabilities don't sound like they would be mutually exclusive. Thanks for the ping, appreciate it! :)

[...]

Issue #87, opened 2020-08-01
z3-sys fails to build on Windows?


zkat:
Hey there! I'm trying to play around with this crate on Windows, and z3-sys is failing to build when I try to do a static build. I'm super new to this and I don't really understand the build errors. Does anyone have any idea what I can do to get this working?

<details>
<summary>Error output</summary>
<pre>
error: failed to run custom build command for `z3-sys v0.6.2`

Caused by:
  process didn't exit successfully: `C:\Users\zkat\src\orogene\target\release\build\z3-sys-6bf7b8d11a33477d\build-script-build` (exit code: 101)
--- stdout
running: "cmake" "C:\\Users\\zkat\\.cargo\\registry\\src\\github.com-1ecc6299db9ec823\\z3-sys-0.6.2\\z3" "-G" "Visual Studio 16 2019" "-Thost=x64" "-Ax64" "-DZ3_BUILD_LIBZ3_SHARED=false" "-DZ3_BUILD_EXECUTABLE=false" "-DZ3_BUILD_TEST_EXECUTABLES=false" "-DCMAKE_INSTALL_PREFIX=C:\\Users\\zkat\\src\\orogene\\target\\release\\build\\z3-sys-efa3561d0d5a5b66\\out" "-DCMAKE_C_FLAGS= -nologo -MT -Brepro" "-DCMAKE_C_FLAGS_RELEASE= -nologo -MT -Brepro" "-DCMAKE_CXX_FLAGS= -DWIN32 -D_WINDOWS -nologo -MT -Brepro" "-DCMAKE_CXX_FLAGS_RELEASE= -DWIN32 -D_WINDOWS -nologo -MT -Brepro" "-DCMAKE_ASM_FLAGS= -nologo -MT -Brepro" "-DCMAKE_ASM_FLAGS_RELEASE= -nologo -MT -Brepro" "-DCMAKE_BUILD_TYPE=Release"
-- Selecting Windows SDK version 10.0.18362.0 to target Windows 10.0.19041.
-- The CXX compiler identification is MSVC 19.25.28614.0
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/BuildTools/VC/Tools/MSVC/14.25.28610/bin/Hostx64/x64/cl.exe
[...]

PR #193, opened 2022-05-10
Use precompiled z3 libs by default

Trolldemorted:
Fixes https://github.com/prove-rs/z3.rs/issues/192 https://github.com/prove-rs/z3.rs/issues/183 https://github.com/prove-rs/z3.rs/issues/87 https://github.com/prove-rs/z3.rs/issues/224 https://github.com/prove-rs/z3.rs/issues/212 and most likely https://github.com/prove-rs/z3.rs/issues/172

This PR tries to solve a number of rough edges of z3-sys mentioned in https://github.com/prove-rs/z3.rs/issues/192:

#### Structural changes
- Changes the default build to link against libraries downloaded from https://github.com/Z3Prover/z3/releases.
- Move bindgen out of z3-sys

#### CI changes
- Add Windows builds
- Run tests on default, dynamic and static builds for all platforms where it works
- Stop ignoring errors in CI

#### QOL changes
- Default installation needs no manual steps and yet does not rebuild z3
- z3-sys with the default features is now completely dependency free (except for z3 itself, ureq and zip) 🎉
- Switching to more coarse conditional compilation in build.rs to increased the readability in my opinion

#### Negative aspects
- So far I threw x86 out. Do you know whether anyone is still using it?
- ~`cargo test` does not execute doc examples because they fail due to a cargo bug: https://github.com/rust-lang/cargo/issues/8531~
- Version bumps of z3 must be done in multiple files, and the bindgen output must match the version

What do you think? I didn't touch the documentation yet, because I wanted to hear your opinion on the technical side of things first.


Bickio:
I was unable to build using latest master and the commit in this PR solved my issue. +1 for merging this change

.github/workflows/rust.yml
```diff
@@ -8,60 +8,47 @@ on:
...
       matrix:
-        build: [linux, macos, windows]
-        include:
-          - build: linux
-            os: ubuntu-latest
-          - build: macos
-            os: macos-latest
-          - build: windows
-            os: windows-latest
+        os: [ubuntu-latest, windows-latest]
```
waywardmonkeys, line 18: We would need macOS coverage here as well.

waywardmonkeys (changes requested):

waywardmonkeys:
The issues with compile times and version fragility have been bothering me a lot as well. I started to wonder "Can we fix the compile time of Z3 itself?" and took a look, but it ends up that that is hard to fix as well.

I think we'll have to sort out a way to have the doc tests run. I've been hoping to add more of those and get rid of some of the standalone tests, so that the docs are more useful.

This PR is using bindgen 0.59, but should move to 0.60, to use the current version (and match what is on the head branch of z3-sys).

Using system-provided Z3 is still problematic in the near future as that is 4.8.x and we'll be wanting to move the bindings forward to using more recent versions of Z3 and supporting the newer APIs that have been added. But that's separate from this, except that having this use the pre-compiled binaries makes the compilation times acceptable.

I'd like to see this move forward for sure.

What're the next steps and is there anything I can do to help?

waywardmonkeys:
It might be worth cherry-picking the parts to use pre-compiled libs before making changes to the bindgen stuff or at least separately from it?

Or if they're too inter-related and both are needed at once, then we just leave this as it is.

waywardmonkeys:
> I was unable to build using latest master and the commit in this PR solved my issue. +1 for merging this change

What build error are you seeing with the latest `master`?

Bickio:
> What build error are you seeing with the latest master?

Linux Mint 20.2
```
  wrapper.h:1:10: fatal error: 'z3.h' file not found
  wrapper.h:1:10: fatal error: 'z3.h' file not found, err: true
  thread 'main' panicked at 'Unable to generate bindings: ()', /home/bickio/.cargo/registry/src/github.com-1ecc6299db9ec823/z3-sys-0.7.1/build.rs:33:14
```
Can provide the full backtrace if required

waywardmonkeys:
Thanks, @Bickio 

I'm looking at the original changes in this PR and have found a few issues ... I'll try and land something by Monday or Tuesday that addresses all of this, with these changes as a starting point.

waywardmonkeys:
@Bickio I haven't forgotten. Something else came up this week and I'm a bit behind.

Trolldemorted:
>It might be worth cherry-picking the parts to use pre-compiled libs before making changes to the bindgen stuff or at least separately from it?

I guess it depends on how early you want to ship the next release, for my personal stuff I can continue with my fork until something like this lands upstream, so I don't have overly urgent time pressure.

.github/workflows/rust.yml
Trolldemorted: I have never used macos in my life, do you know how one would install z3 system-wide there?

Related: I just noticed we don't have any guarantee that the preinstalled z3 on the linux machines have a compatible z3 version installed. Can/should we uninstall the preinstalled z3 and install the correct version for `dynamic-link-z3` on linux?

Trolldemorted:
Any chance we can get this (or a similar implementation) upstream at some point? 😇

z3-sys/build.rs
```diff
@@ -1,49 +1,78 @@
...
+        .unwrap()
+        .wait()
+        .unwrap();
+
+    #[cfg(not(target_os = "windows"))]
+    let tar = std::process::Command::new("unzip")
+        .current_dir(archive_path.parent().unwrap())
+        .arg(&archive_path.to_str().unwrap())
+        .spawn()
+        .unwrap()
```
connorskees, line 57: The error message here can be a bit confusing if one doesn't have the `unzip` binary installed, as the error message shown is just that a file somewhere is missing.

connorskees:
My use case is running this library on a somewhat under-powered server, where when trying to build z3 LLVM runs for several minutes until it fails due to OOM. Using this branch lets my project build successfully.

z3-sys/build.rs
Trolldemorted: would it be ok to introduce a new dev dep that handles that stuff?

Trolldemorted:
Done some changes:
- rebased onto master
- added CI for static and default macos builds
- moved bindgen out of z3-sys to reduce z3 rebuilds
- moved the bindgen output to target-specific folders, since the bindgen output on different targets is not equal
- replaced curl, unzip and tar with the `ureq` and `zip` crates
- halved CI runtime by no longer building z3 twice
- "fix" doctests by working around https://github.com/rust-lang/cargo/issues/8531

I have no idea how builds and CI tests for native z3 installations on macos would look like, because master does not have them either. Since you had to disable tests for native z3 installations for ubuntu on master, I sincerely recommend scrapping that feature completely. We will never get it safe, and at some point it is going to hurt someone.

@waywardmonkeys PTAL, the current state of master is disadvantageous :)

waywardmonkeys (changes requested):
Right now, this is doing too many things and too big.

I'd like to only review changes for now that download binaries and then use them.

I don't want to split out `bindgen` and such for now ... and other things maybe see in future / subsequent PRs.

stefnotch:
@waywardmonkeys I would hate to see this PR go to waste. Is there a reasonable way forward for this?

PR #196, opened 2022-06-03
added binding for Z3_mk_solver_from_tactic 

AndesVL:


waywardmonkeys:
Can you remove the `as_f64` change from this PR so that it only has in it the single required commit?

AndesVL:
done!

PR #198, opened 2022-06-05
added bindings for z3_param_descrs and related functions

AndesVL:
Introduced a new struct: ParamDescrs which acts as a wrapper for Z3_param_descrs.
These can be constructed from Tactic and Solver.
Provides a typesafe wrapper for *most* related functions (in doing so introduced a conversion from Symbol to Z3_Symbol.


toolCHAINZ:
I'm going to close this for inactivity, as the codebase has shifted quite a bit under it, but I do think this would be useful to have. Maybe something like this could be used as an alternative to something like #393.

PR #199, opened 2022-06-05
added bindings for global params

AndesVL:


waywardmonkeys:
I had to revert this, so if you could fix the failing tests and re-submit a PR, that'd be great.

Sorry for landing it and causing the extra work rather than catching it here in this PR.

Shatur:
@waywardmonkeys can we have these functions without a test since they modify global params?
It sometimes needed to adjust the solver.

Shatur:
@waywardmonkeys opened #242.

Issue #200, opened 2022-07-19
z3 v4.8.12 has memory leaks


Trolldemorted:
Do you have plans to upgrade?

waywardmonkeys:
Let's see what happens if I update to 4.9.x ...

PR #203, opened 2022-07-21
Update static z3 from 4.8.12 to 4.10.1.

waywardmonkeys:


waywardmonkeys:
This seems to work for me locally (the tests pass) and it may help out with #200.

waywardmonkeys:
Wait a day and now I get to update to 4.10.1 instead of 4.9.1 :)

dariusf:
FWIW, I've been running this branch locally for a while and it works well. z3 4.8 is somehow much slower for my use case and the upgrade fixes this.

genos:
Looks like Z3 is now up to 4.12.2; is it worth updating even further?

PR #208, opened 2022-07-27
Deny warnings in ci

SeeSpring:


waywardmonkeys:
Thanks. I think if someone needs the context from here, they can just get it from the underlying sort instead in the future.

PR #209, opened 2022-07-27
Fix clippy errors

SeeSpring:
Based on #178 to avoid merge conflicts. I can separate it if you want


waywardmonkeys:
I had been doing a version of the change to `Ast::new`, but I was also hoisting the Z3 mutex up higher in each thing so that it covered more ... but I'll do that as a follow-on patch to this, I think.

I also was thinking of renaming these `new` functions to be `wrap` since that's what they're actually doing ... but that again can be a subsequent patch.

PR #216, opened 2022-12-14
Make it buildable for wasm with emscripten

### LINKED ISSUE #215: Wasm build
Hi, and thanks for this great library! Is it possible to use it with web assembly? Either `wasm32-unknown-unknown` or `wasm32-wasi`? My naive try failed with `Can not find z3.h` error, but I'm not sure if something is wrong on my side. It seems [z3 itself supports webassembly](https://github.com/Z3Prover/z3#web-assembly--typescript--javascript) so I wonder if there is a way to tweak this library and make it work with wasm.

HKalbasi:
fix #215

Surprisingly, it just needed to use emscripten's sysroot with bindgen. For `wasm32-wasi` target more work is needed, as it doesn't support c++ exceptions yet, but I think `wasm32-unknown-emscripten` is enough for most of use cases (including my use case).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -33,6 +33,21 @@ jobs:
       # See above.
       if: ${{ success() || failure() }}
 
+  build_on_wasm:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install emscripten
+      run: |
+        git clone https://github.com/emscripten-core/emsdk.git
+        cd emsdk
+        git pull
+        ./emsdk install latest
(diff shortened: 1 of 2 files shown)


waywardmonkeys (changes requested):
How does this build Z3 itself and how does that work?

Could you remove the change to `.gitignore` and do that as a separate thing?

HKalbasi:
`cmake` crate is aware that it should use `emcmake` when target is `wasm32-unknown-emscripten` and z3 is ok with emscripten (they use it for js builds). So everything works automatically.

I removed `.gitignore` and added a CI check for wasm build.

PR #222, opened 2023-02-21
rename context args to ctx

Pat-Lafon:
Hello, I would like to suggest a trivial QoL change.

Most functions take a context parameter as `ctx: &'ctx Context`. There are only a couple that take `contex: &'ctx Context`. There is no semantic difference here, however, this does mean that in a few cases an editor will autocomplete a method/function call with a `context` variable name instead of a `ctx`. This is mildly annoying as a user.

This is no doubt a limitation of my autocomplete tooling(the RA extension of VSCode). Despite that, given that the use of `ctx` as an argument name is so prevalent in this crate and its test cases, I am suggesting that the naming be consistent and have done my best to grep for these inconsistencies.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -149,13 +149,13 @@ macro_rules! varop {
     ) => {
         $(
             $( #[ $attr ] )*
-            pub fn $f(context: &'ctx Context, values: &[&Self]) -> $retty {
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
Fully agree, thanks!

PR #223, opened 2023-02-23
add regexp support

laszlo-nodebee:
This PR adds a new `Regexp` struct to `ast` along with the corresponding operations.


waywardmonkeys:
Thanks, I'm going to merge this.

But to me, this just shows what I don't like about any of the existing Z3 API in this crate with too much boilerplate and use of `ctx` parameters everywhere. Just food for thought...

PR #228, opened 2023-03-21
Update dependencies and appease Clippy

genos:
First of all, thanks _very_ much for this project! It makes it very straightforward to reach for Z3 from Rust.

If you don't mind, this PR makes two changes, both in the interest of appeasing Clippy:

1.  bump dependencies in `z3-sys/Cargo.toml` to ensure a newer `nom` version is used, ameliorating a Clippy's objection `warning: the following packages contain code that will be rejected by a future version of Rust: nom v5.1.2`
2. other small changes in `z3` to appease Clippy & silence warnings

Thanks!


genos:
I've also taken the liberty of bumping Z3 to 4.12.1 and fixing some tests.

PR #230, opened 2023-03-31
Expensive copy of a solver

TomMD:
Z3 has no native way to copy a solver, commonly useful for BFS style tree traversals of contexts of execution.  Sometimes solver push/pop suffices and sometimes you just need to copy.  This code uses the [same operations Leonardo suggested](https://stackoverflow.com/questions/16516337/copying-z3-solver) on the topic to cope with Z3's lack.

### DIFF
--- a/z3/src/solver.rs
+++ b/z3/src/solver.rs
@@ -314,3 +314,25 @@ impl<'ctx> Drop for Solver<'ctx> {
         unsafe { Z3_solver_dec_ref(self.ctx.z3_ctx, self.z3_slv) };
     }
 }
+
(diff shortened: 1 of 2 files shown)


waywardmonkeys (changes requested):
This needs a test of some sort. If there's a pointer to the place where this is described as the right approach upstream, it would be good to have it as a comment for this function as well so that the provenance of the approach is clear.

waywardmonkeys:
Oops, I see that you linked the place where he said it here in the PR. Please also put it as a comment in the code.

waywardmonkeys:
Thanks!

PR #218, opened 2022-12-15
Fix ci for wasm build

HKalbasi:

PR #231, opened 2023-04-01
Move clippy into CI

Pat-Lafon:
With #228 and previously #217, a couple of pr's have been accepted to address clippy suggestions. Given this, it seems reasonable to move clippy into the github actions workflow.

This pr:
- Adds an additional, parallel job to the workflow to run clippy across the whole workspace.
- Picks some low hanging ci performance fruit by using `CARGO_REGISTRIES_CRATES_IO_PROTOCOL: sparse` and not statically building `z3` twice(As I later found mentioned in https://github.com/prove-rs/z3.rs/pull/207#issuecomment-1196265484).
- Fixes the last of clippy's suggestions in various build/test files.

Depending on the status of #207, it would be faster to run clippy without `static-link-z3`. Sometimes people also move `cargo fmt` into the ci.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   CARGO_INCREMENTAL: 0
+  CARGO_REGISTRIES_CRATES_IO_PROTOCOL: sparse
   RUSTFLAGS: "-D warnings"
 
 jobs:
@@ -39,12 +40,6 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: recursive
-    - name: Install LLVM and Clang # required for bindgen to work, see https://github.com/rust-lang/rust-bindgen/issues/1797
-      uses: KyleMayes/install-llvm-action@v1
-      if: matrix.os == 'windows-latest'
-      with:
-        version: "11.0"
-        directory: ${{ runner.temp }}/llvm
     - name: Install emscripten
       run: |
         cd ~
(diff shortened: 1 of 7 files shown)


.github/workflows/rust.yml
```diff
@@ -39,12 +39,6 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: recursive
-    - name: Install LLVM and Clang # required for bindgen to work, see https://github.com/rust-lang/rust-bindgen/issues/1797
-      uses: KyleMayes/install-llvm-action@v1
-      if: matrix.os == 'windows-latest'
-      with:
-        version: "11.0"
-        directory: ${{ runner.temp }}/llvm
```
waywardmonkeys, line 47: This isn't needed anymore?

waywardmonkeys (changes requested):
Can you rebase the commit for running clippy in parallel into the commit where you added the config to run clippy in CI?

.github/workflows/rust.yml
Pat-Lafon: We can ping @HKalbasi who added it in #218 to confirm but this was caught by the Github Actions linter.

It identifies that the `if: matrix.os == 'windows-latest'` will never be true so this step does not fire(this is verified by looking at the most recent run for this repository https://github.com/prove-rs/z3.rs/actions/runs/4537746078/jobs/7995914848). This condition tries to access this part of the context `matrix.os` that is never set in this specific job.

Additionally, this fix (afaik) is only needed when running on windows, this job uses `ubuntu-latest` which is why it this works despite the step never firing. In looking back at the original commit https://github.com/prove-rs/z3.rs/pull/218/commits/a4fb9dd195d8f1933464e677b13e33ddf8aef064, I'm unsure that this was ever needed in this specific job.

waywardmonkeys: You're correct! This isn't needed here since this doesn't run on Windows.

waywardmonkeys (approved):

Issue #221, opened 2023-02-12
Iterator for a solver


onthestairs:
Hi, firstly, thank you very much for this libary - it's great! 

I'm not sure if this is a generic rust question, or something specific to this library. I am trying to create an 'iterator' of solutions from z3, so that I can access the solutions 'lazily' and in a composable way (For example taking only the first n solutions with `.take(n)`). However, I can't figure out the ergonomics of it, and I'm not sure if it will even be possible to make the lifetimes work. 

Here is a version of what I would like to do which doesn't use iterators, and just returns a `Vec` of all the solutions:
```rust
use z3::ast;
use z3::ast::Ast;
use z3::Config;
use z3::Context;
use z3::SatResult;
use z3::Solver;

fn solver(target: i64) -> Vec<(i64, i64)> {
    let cfg = Config::new();
    let ctx = Context::new(&cfg);
    let solver = Solver::new(&ctx);

    let x = ast::Int::new_const(&ctx, "x");
    let y = ast::Int::new_const(&ctx, "y");
    let n = ast::Int::from_i64(&ctx, target);
    let equals = ast::Int::add(&ctx, &[&x, &y])._eq(&n);
    solver.assert(&x.lt(&n));
    solver.assert(&y.lt(&n));
    solver.assert(&equals);

    let mut solutions = vec![];
    while solver.check() == SatResult::Sat {
        let model = solver.get_model().unwrap();
        let x_val = model.eval(&x, true).unwrap();
        let y_val = model.eval(&y, true).unwrap();
        let solution = (x_val.as_i64().unwrap(), y_val.as_i64().unwrap());
        solutions.push(solution);

        let is_the_solution = ast::Bool::and(&ctx, &[&x._eq(&x_val), &y._eq(&y_val)]);
        solver.assert(&is_the_solution.not());
    }
    return solutions;
}

fn main() {
    for (x, y) in solver(10) {
        println!("x={}, y={}", x, y)
    }
}
```

Here is an attempt to figure out what it 'might' look like. But I cannot arrange it to satisfy the lifetime checker:
```rust
use z3::ast;
use z3::ast::Ast;
use z3::Config;
use z3::Context;
use z3::SatResult;
use z3::Solver;

struct AddSolver<'a> {
    ctx: Context,
    solver: Solver<'a>,
    x: ast::Int<'a>,
    y: ast::Int<'a>,
}

impl<'a> Iterator for AddSolver<'a> {
    type Item = (i64, i64);

[...]

PR #232, opened 2023-04-04
Implement Iterator over a Model

Pat-Lafon:
One nice abstraction that the python bindings implement is to be able to iterate over a `Model`. One can then use the features added in #221 to get the interpretations that the `Model` assigns for these declarations.

If this is worth adding, the implementation comes from https://z3prover.github.io/api/html/classz3py_1_1_model_ref.html#a7890b7c9bc70cf2a26a343c22d2c8367 where they hook into Python's `__len__` and `__getitem__` to extract all consts and funcs from a model when using a for loop. I think the only difference here is that for Rust we need to manage a struct that does the iteration, here `ModelIter`(akin to https://doc.rust-lang.org/std/slice/struct.Iter.html) which contains the current index into the model and the total length.

The testcase is related to #221.

### DIFF
--- a/z3/src/model.rs
+++ b/z3/src/model.rs
@@ -126,6 +126,17 @@ impl<'ctx> Model<'ctx> {
             None
         }
     }
+
+    fn len(&self) -> u32 {
+        unsafe {
+            Z3_model_get_num_consts(self.ctx.z3_ctx, self.z3_mdl)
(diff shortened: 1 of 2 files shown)


waywardmonkeys (approved):

waywardmonkeys:
Thanks!

PR #234, opened 2023-04-08
Aligning `apply` methods of FuncDecl and RecFuncDecl

Pat-Lafon:
Given that both structs(`FuncDecl` and `RecFuncDecl`) provide the same functionality modulo `add_def`, it's kind of unfortunate that the signatures for `apply` uses a different type for the `args`.

https://github.com/prove-rs/z3.rs/blob/691ac9782f361969e87ac745a3d9a6816593741a/z3/src/func_decl.rs#L62

vs 

https://github.com/prove-rs/z3.rs/blob/691ac9782f361969e87ac745a3d9a6816593741a/z3/src/rec_func_decl.rs#L117

In this case it's slightly better to pass a `&dyn Ast` instead of a `&Dynamic` because the latter may require creating a temporary that is only used by reference when creating an array to pass as `args`(which can make some code a bit awkward).

I've made this short change in the first commit and removed the `into` conversions from the tests which are now ambiguous.

I've also thought of a more aggressive change which implements [`Deref`](https://doc.rust-lang.org/std/ops/trait.Deref.html) for `RecFuncDecl`. With this, the compiler can use auto-deref magic to allow the user to call `FuncDecl` methods on a `RecFuncDecl`. Note how this change does not require any changes to the tests. With this, I'm making an assumption that the implementation of both structs is the same. In general, this kind of change would bring the underlying implementation of this functionality more in line with the other language bindings. This is implemented in the second commit.

It could also be reasonable to just make `RecFuncDecl` a wrapper struct containing a `FuncDecl` which would simplifying all of it's impls greatly.

If this issue makes sense to fix, I'm not particularly attached to any of the solutions here and would be happy to adjust this pr as needed(only the first commit, only the second commit, modifying `RecFuncDecl` to be a wrapper around `FuncDecl`, or something else).

### DIFF
--- a/z3/src/lib.rs
+++ b/z3/src/lib.rs
@@ -181,6 +181,8 @@ pub struct FuncEntry<'ctx> {
 /// the sort (i.e., type) of each of its arguments. This is the function declaration type
 /// you should use if you want to add a definition to your function, recursive or not.
 ///
+/// This struct can dereference into a [`FuncDecl`] to access its methods.
+///
 /// # See also:
 ///
 /// - [`RecFuncDecl::add_def`]
--- a/z3/src/rec_func_decl.rs
+++ b/z3/src/rec_func_decl.rs
@@ -3,8 +3,9 @@ use ast::Ast;
 use std::convert::TryInto;
 use std::ffi::CStr;
 use std::fmt;
+use std::ops::Deref;
 use z3_sys::*;
-use {Context, RecFuncDecl, Sort, Symbol};
+use {Context, FuncDecl, RecFuncDecl, Sort, Symbol};
 
 impl<'ctx> RecFuncDecl<'ctx> {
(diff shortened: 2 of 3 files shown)


waywardmonkeys:
This all sounds reasonable to me at first read ... can @TomMD or someone else who used things recently chime in with their thoughts?

Pat-Lafon:
While I'm at it, I'll make the suggestion to change `add_def`'s signature as well. https://github.com/prove-rs/z3.rs/blob/691ac9782f361969e87ac745a3d9a6816593741a/z3/src/rec_func_decl.rs#L74 The `args` parameter has a similar argument as apply.

It's interesting that `body` uses `&impl Ast`. There are small difference between this and `&dyn Ast` in terms of code dispatch but I don't think there would be any noticeable difference here to the user. I've changed that to `&dyn Ast` for uniformity but can change it back if needed.

waywardmonkeys:
Let's just do this.

PR #236, opened 2023-04-12
Fix smt based tests

Pat-Lafon:
In playing around with enabling the error_handler, I noticed that the tests for #226 were actually incorrect but there was no error checking so this was just ignored.(Whoops, my bad)

Maybe the api should be changed to avoid this? Something akin to https://github.com/prove-rs/z3.rs/blob/691ac9782f361969e87ac745a3d9a6816593741a/z3/src/tactic.rs#L207-L211. In general, I think there should be a better story for integrating with Z3's errors.


waywardmonkeys:
Agree that we should do something better about the error integration!

Issue #235, opened 2023-04-12
Symbol::as_z3_symbol can return Null Pointer


Pat-Lafon:
It seem like the integer conversion for Symbol does not check that the integer is within the valid bounds that `Z3_mk_int_symbol` expects(0 <= 2^30 -1). If you try to create a symbol with `u32::MAX` for example then any of the functions that use a symbol via `as_z3_symbol` will be using a null pointer.

I'm not sure what bad things can happen because of this, I think some of the usual cases do get caught or just have other operations return null as well.

```rust
#[test]
fn test() {
    let cfg = crate::Config::new();
    let ctx = Context::new(&cfg);
    println!("{}", crate::ast::Bool::new_const(&ctx, u32::MAX).to_string());
}
```

Pat-Lafon:
In doing some more digging, I see that Z3 tries to warn the user about this but in #92 we disabled the error handler and don't otherwise try to detect this situation.

PR #242, opened 2023-07-13
Add bindings for global params

Shatur:
Resubmit of #199.

### DIFF
--- a/z3/src/context.rs
+++ b/z3/src/context.rs
@@ -1,3 +1,4 @@
+use std::ffi::CString;
 use z3_sys::*;
 use Config;
 use Context;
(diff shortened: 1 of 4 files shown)


Pat-Lafon:
Hi, I was curious, what happens when you call https://z3prover.github.io/api/html/group__capi.html#ga9162ac23461bc67d1a0cde7c71f53111 with some bad string(Out of range chars or incorrect use of delimitation)?

Does it panic or just silently fail? Do we expect these bindings to enforce the precondition or is it up to Z3?

Shatur:
> with some bad string

But we pass `str`, Rust already ensures that the string is valid. Unless you use `unsafe` :smile:
The same code already used in Z3 for `Config`.

Shatur:
A bit off-topic question, but do you know a way to set "global" parameters for specific context?
I tried setting `sat.euf` via `Params` or via `Z3_update_param_value`, but only global params have any effect.

Pat-Lafon:
Sorry, I don't know too much about this. Just fyi, if this is all through the Rust bindings, the Z3 error handler gets disabled so it might be silently failing when you are trying to set the parameter. https://github.com/prove-rs/z3.rs/issues/235#issuecomment-1505779412 If that's your issue, then re-enabling it should shed light on the situation.

Shatur:
@Pat-Lafon figured it out:
```rust
        let tactic = Tactic::new(&ctx, "sat");
        let solver = tactic.solver();
        let mut params = Params::new(&ctx);
        params.set_bool("sat.euf", true);
        solver.set_params(&params);
```
But it's unrelated to the PR, I think it's ready for merge.

Pat-Lafon:
Maybe worth pinging @waywardmonkeys to put this on his radar though I'm not sure how much time he has these days.

waywardmonkeys:
Ready for merge? Here we go!

Shatur:
@waywardmonkeys could you draft a new release?

waywardmonkeys:
Let's see in 24 hours ... I'm away from home, so I won't be able to tomorrow. I just pushed some other changes that I had hanging around as well, so CI is backed up for a while.

PR #248, opened 2023-09-10
Use system-shipped z3 libraries by pkg_config

wtdcode:
Probably a bit more elegant way to use precompiled binaries compared to #193 by utilizing `pkg-config`, which is ubiquitous of all *nix platforms.

My usecase is to use homebrew installed z3 library.

Changes:

- Detect dynamic/static precompiled libraries and use them with `pkg_config`
- Add CI for macos using precompile libraries from homebrew. 
- Instead of manually encoded `z3/src/api/z3.h`, use the universal `z3.h` with corresponding include directories added, which should be a more general approach.

No break changes are introduced in theory and should be safe to go.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -20,11 +20,23 @@ jobs:
       run: cargo fmt -- --check
 
   build:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        build: [linux, macos]
+        include:
+          - build: linux
+            os: ubuntu-latest
+          - build: macos
+            os: macos-latest
+    runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v3
     - name: Install Z3
+      if: matrix.os == 'ubuntu-latest'
       run: sudo apt-get install libz3-dev
+    - name: Install Z3
+      if: matrix.os == 'macos-latest'
+      run: brew install z3
     - name: Build
       run: cargo build -vv --all
       # XXX: Ubuntu's Z3 package seems to be missing some symbols, like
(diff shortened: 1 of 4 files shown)


waywardmonkeys:
There is also now #249 which differs some from this ...

wtdcode:
> There is also now #249 which differs some from this ...

It’s totally different because it will use dynamic libraries from distributions like homebrew, Debian etc.

waywardmonkeys:
Okay, so now we have `vcpkg` support. What scenarios does having  this help us in? We already build out of the box with syste, provided z3 on macOS (via homebrew) and with the Ubuntu package. Is it that you could build your own and have it elsewhere?

wtdcode:
> Okay, so now we have `vcpkg` support. What scenarios does having this help us in? We already build out of the box with syste, provided z3 on macOS (via homebrew) and with the Ubuntu package. Is it that you could build your own and have it elsewhere?

I can't understand how `vcpkg` helps here.

poscat0x04:
Can confirm this works on fedora without manually setting the environment variable!

poscat0x04:
@waywardmonkeys The problem is that bindgen is not finding the header files when it should be because the include path is not set up correctly. Most linux distributions use pkg-config to specify C library metadata such as include path, linker flags etc.

poscat0x04:
I've fixed the conflict, here is a patch: [resolve-conflict.patch](https://github.com/prove-rs/z3.rs/files/13256032/resolve-conflict.patch)

waywardmonkeys:
I will get this landed later today when I am back home.

poscat0x04:
politely pings @waywardmonkeys

waywardmonkeys:
I've done a fresh update of this PR in #277 and will land it once it passes CI.

Thanks for the initial work!

PR #250, opened 2023-10-11
is_const_array and Distinct with &[impl Borrow<Self>]

Pat-Lafon:
- Distinct currently takes `&[&Self]` for which I would like to do a similar api relaxing like in #246.
- With the new Rust version, there is a minor clippy change.
- Python has this nice `is_const_array(a)` function which I would like to propose here.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -264,14 +264,17 @@ pub trait Ast<'ctx>: fmt::Debug {
     /// `Ast`s being compared must all be the same type.
     //
     // Note that we can't use the varop! macro because of the `pub` keyword on it
-    fn distinct(ctx: &'ctx Context, values: &[&Self]) -> Bool<'ctx>
+    fn distinct(ctx: &'ctx Context, values: &[impl Borrow<Self>]) -> Bool<'ctx>
     where
         Self: Sized,
     {
         unsafe {
             Bool::wrap(ctx, {
                 assert!(values.len() <= 0xffffffff);
-                let values: Vec<Z3_ast> = values.iter().map(|nodes| nodes.get_z3_ast()).collect();
+                let values: Vec<Z3_ast> = values
+                    .iter()
+                    .map(|nodes| nodes.borrow().get_z3_ast())
+                    .collect();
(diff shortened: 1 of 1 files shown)


z3/src/ast.rs
```diff
@@ -1525,6 +1528,26 @@ impl<'ctx> Array<'ctx> {
             })
         }
     }
+
+    /// Returns true if the array is a const array(a.is_const_array() <=> exists v, forall i. select(a, i) == v)
```
waywardmonkeys, line 1532: Is this correct?

waywardmonkeys (changes requested):

z3/src/ast.rs
Pat-Lafon: That is coming from https://github.com/prove-rs/z3.rs/blob/3c9766336297769467be0ddcbce0376b58a344f3/z3-sys/src/lib.rs#L520-L522. So I believe so? Maybe only a single implication now that I think of it because I'm not sure if a nested store gets collapsed. (store (store ((as const (Array Int Int)) 9) 0 1281) 0 9)

waywardmonkeys: Sorry, I had commented while doing something else and from my phone away from home.

My concern was actually about the formatting and if that should be in the summary line at the top of the comment.

Should also have a blank line before the example header.

z3/tests/lib.rs
```diff
@@ -1221,7 +1221,7 @@ fn test_tactic_fail() {
 
     let tactic = Tactic::new(&ctx, "fail");
     let apply_results = tactic.apply(&goal, Some(&params));
-    assert!(matches!(apply_results, Err(_)));
+    assert!(apply_results.is_err());
```
waywardmonkeys, line 1224: I've landed this separately (also done by another PR). This commit can be removed from this PR.

waywardmonkeys:
Could you squash the various commits for `is_const_array` into a single commit? After removing the clippy fix (already on `master`), that would leave one commit for the `Distinct` change and one for `is_const_array`.

PR #251, opened 2023-10-13
Add vcpkg support and corresponding CI.

TheVeryDarkness:
Support using [vcpkg-rs](https://github.com/mcgoo/vcpkg-rs) to manage z3 library.

Added a non-default feature vcpkg.

However, vcpkg-rs does not support wasm32 target currently. I created a pull request there at [mcgoo/vcpkg-rs#53](https://github.com/mcgoo/vcpkg-rs/pull/53).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,6 +12,13 @@ env:
   RUSTFLAGS: "-D warnings"
 
 jobs:
+  check-formatting:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Check formatting
+      run: cargo fmt -- --check
+
   build:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -94,6 +94,46 @@ jobs:
         run: cargo test -vv --features static-link-z3
       - name: Test `z3` with statically linked Z3 and `arbitrary-size-numeral` enabled
         run: cargo test --manifest-path z3/Cargo.toml -vv --features 'static-link-z3 arbitrary-size-numeral'
+  build_with_vcpkg_installed_z3:
+    strategy:
+      matrix:
+        build: [linux, macos, windows]
+        config:
+          - os: ubuntu-latest
+            vcpkg_triplet: x64-linux
+          - os: macos-latest
+            vcpkg_triplet: x64-osx
+          - os: windows-latest
+            vcpkg_triplet: x64-windows-static-md
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Install LLVM and Clang # required for bindgen to work, see https://github.com/rust-lang/rust-bindgen/issues/1797
+        uses: KyleMayes/install-llvm-action@v1
+        if: matrix.os == 'windows-latest'
+        with:
+          version: "11.0"
+          directory: ${{ runner.temp }}/llvm
+      - name: Set LIBCLANG_PATH
+        run: echo "LIBCLANG_PATH=$((gcm clang).source -replace "clang.exe")" >> $env:GITHUB_ENV
+        if: matrix.os == 'windows-latest'
+      - name: vcpkg build z3
+        uses: johnwason/vcpkg-action@v5
+        id: vcpkg
+        with:
+          pkgs: z3
+          triplet: ${{ matrix.config.vcpkg_triplet }}
+          cache-key: ${{ matrix.config.os }}
+          revision: master
+          token: ${{ github.token }}
+      - name: Build `z3-sys` and `z3` with vcpkg installed Z3
+        run: cargo build -vv --features vcpkg
+      - name: Test `z3-sys` and `z3` with vcpkg installed Z3
+        run: cargo test -vv --features vcpkg
+      - name: Test `z3` with vcpkg installed Z3 and `arbitrary-size-numeral` enabled
+        run: cargo test --manifest-path z3/Cargo.toml -vv --features 'vcpkg arbitrary-size-numeral'
 
   run_clippy:
     runs-on: ubuntu-latest
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -97,15 +97,20 @@ jobs:
   build_with_vcpkg_installed_z3:
     strategy:
       matrix:
-        build: [linux, macos, windows]
+        build: [windows] # [linux, macos, windows]
         include:
-          - os: ubuntu-latest
-            vcpkg_triplet: x64-linux
-          - os: macos-latest
-            vcpkg_triplet: x64-osx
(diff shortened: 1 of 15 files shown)


z3-sys/build.rs
```diff
@@ -17,6 +19,33 @@ fn main() {
...
+    generate_binding(&header);
+}
+
+#[cfg(feature = "vcpkg")]
+fn main() {
+    let lib = vcpkg::Config::new()
+        .emit_includes(true)
+        .find_package("z3")
+        .unwrap();
+    let found_header = lib.include_paths.iter().any(|include| {
```
waywardmonkeys, line 32: Does this actually have more than one include path present?

z3/tests/lib.rs
```diff
@@ -1221,7 +1221,7 @@ fn test_tactic_fail() {
 
     let tactic = Tactic::new(&ctx, "fail");
     let apply_results = tactic.apply(&goal, Some(&params));
-    assert!(matches!(apply_results, Err(_)));
+    assert!(apply_results.is_err());
```
waywardmonkeys, line 1224: I've landed this and some other clippy-related changes separately, so this commit can be removed.

z3-sys/build.rs
```diff
@@ -31,16 +60,26 @@ fn main() {
...
             .parse_callbacks(Box::new(bindgen::CargoCallbacks))
             .generate_comments(false)
             .rustified_enum(format!("Z3_{}", x))
             .allowlist_type(format!("Z3_{}", x));
-        if env::var("TARGET").unwrap() == "wasm32-unknown-emscripten" {
-            enum_bindings = enum_bindings.clang_arg(format!(
-                "--sysroot={}/upstream/emscripten/cache/sysroot",
-                env::var("EMSDK").expect("$EMSDK env var missing. Is emscripten installed?")
-            ));
+        let target = env::var("TARGET").unwrap();
```
waywardmonkeys, line 68: I'd rather have the emscripten / webassembly changes as a separate PR. I'm less sure about what's happening here and it isn't directly related to the `vcpkg` work (to me ... my main interest is in using `vcpkg` to improve Windows build support where we don't have an option to use a system-installed Z3).

z3-sys/build.rs
```diff
@@ -17,6 +19,33 @@ fn main() {
     };
     println!("cargo:rerun-if-env-changed={}", Z3_HEADER_VAR);
     println!("cargo:rerun-if-changed={}", header);
+
+    generate_binding(&header);
+}
+
+#[cfg(feature = "vcpkg")]
```
waywardmonkeys, line 26: I'd rather rename `build_z3` to `build_bundled_z3` and then add a new function that uses `vcpkg` and just call it from the existing `main` ...

waywardmonkeys: This would also take care of things like the missing `rerun-if-changed` stuff in the new vcpkg-specific copy of `main`.

waywardmonkeys (changes requested):
We'll also want to do something to add CI coverage of this to make sure it works ... like adding a vcpkg build on Windows in CI.

waywardmonkeys:
Despite all of the above, I do want to say that I'm very happy to see this and thanks so much for starting us down this path!

TheVeryDarkness:
Okay, but I'm afraid that I need some time for it. I may be busy for several days as I'm attending a game.

z3-sys/build.rs
TheVeryDarkness: Normally there will be exactly one. I just put it here in case something changes in the future.

z3-sys/build.rs
```diff
@@ -31,16 +60,26 @@ fn main() {
...
-        if env::var("TARGET").unwrap() == "wasm32-unknown-emscripten" {
-            enum_bindings = enum_bindings.clang_arg(format!(
-                "--sysroot={}/upstream/emscripten/cache/sysroot",
-                env::var("EMSDK").expect("$EMSDK env var missing. Is emscripten installed?")
-            ));
+        let target = env::var("TARGET").unwrap();
+        let wasm32 = target.starts_with("wasm32-unknown");
+        let wasm32_emscripten = target == "wasm32-unknown-emscripten";
+        if wasm32 {
+            let sysroot = env::var("EMSDK")
```
TheVeryDarkness, line 72: It seems that sometimes the cache may be missing for fresh emscripten installation. So I'm not sure whether we should add some notices here.

z3/tests/lib.rs
TheVeryDarkness: I'll merge that into my branch, thanks.

z3-sys/build.rs
TheVeryDarkness: Yes, I've renamed it and added a `rerun-if-changed`. And related env var should have been watched inside vcpkg-rs. Thanks.

z3-sys/build.rs
TheVeryDarkness: Reverted the commit. I'll have more tests and then create a new PR.

.github/workflows/rust.yml
```diff
@@ -94,6 +94,46 @@ jobs:
         run: cargo test -vv --features static-link-z3
       - name: Test `z3` with statically linked Z3 and `arbitrary-size-numeral` enabled
         run: cargo test --manifest-path z3/Cargo.toml -vv --features 'static-link-z3 arbitrary-size-numeral'
+  build_with_vcpkg_installed_z3:
+    strategy:
+      matrix:
+        build: [linux, macos, windows]
```
waywardmonkeys, line 100: For now, let's just do this on Windows since that's where GitHub has it installed.

waywardmonkeys: I saw it failing on Linux and that caused the macOS and Windows builds to get cancelled. (At least in the run that I looked at.)

TheVeryDarkness: I've added an option to disable fail-fast in my branch. Here's the [result](https://github.com/TheVeryDarkness/z3.rs/actions/runs/6561354966/job/17820867395). I guess the output implies that the ABI doesn't match.

TheVeryDarkness: This may be fixed by cleaning the vcpkg build tree.

z3-sys/build.rs
```diff
@@ -1,13 +1,45 @@
 use std::env;
 
+#[cfg(not(feature = "vcpkg"))]
 const Z3_HEADER_VAR: &str = "Z3_SYS_Z3_HEADER";
 
 fn main() {
+    // Feature `vcpkg` is prior to `static-link-z3` as vcpkg-installed z3 is also statically linked.
+
+    #[cfg(not(feature = "vcpkg"))]
     #[cfg(feature = "static-link-z3")]
```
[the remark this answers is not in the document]
TheVeryDarkness: Okay, I see. I may do that tomorrow as it's midnight here. Thanks.

waywardmonkeys:
I'm going to merge this and then see about further / subsequent changes!

waywardmonkeys:
Thank you again for working on this!

PR #252, opened 2023-10-22
Implement `ast::Float::as_f64`

waywardmonkeys:
Fixes https://github.com/Z3Prover/z3/issues/6958

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -626,6 +626,10 @@ impl<'ctx> Float<'ctx> {
             })
         }
     }
+
+    pub fn as_f64(&self) -> f64 {
+        unsafe { Z3_get_numeral_double(self.ctx.z3_ctx, self.z3_ast) }
+    }
 }
 
 impl_ast!(Datatype);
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
@Pat-Lafon Any thoughts on this?

Pat-Lafon:
Simple enough! Though it is kinda weird that the other `.as_*()` functions return an optional. Here sometimes you will get `0.0` because the error facilities are suppressed.

```rust
#[test]
fn test_float() {
    let cfg = Config::new();
    let ctx = Context::new(&cfg);

/*     let f = ast::Float::from_f64(&ctx, 1.0); */
    let f = ast::Float::new_const_double(&ctx, "test");
    assert_eq!(f.as_f64(), 0.0);
}
```

waywardmonkeys:
The underlying API from Z3 differs though which is kind of unfortunate. The C++ binding checks the error code. The C# binding does not. Interestingly (perhaps), other bindings don't seem to call `Z3_get_numeral_double()`.

Maybe I should do so here and return an `Option`? Although it would be nice to have a more ergonomic way to check that error than we do now ... (There's so much un-ergonomic about this code.)

Pat-Lafon:
What do you think would be the more ergonomic solution? Should most of these functions return a result type so the error is not suppressed? (Relying on a lot of `?` everywhere) Runtime assertion panic on a case by case basis?

PR #256, opened 2023-10-23
Always use `num`, remove `arbitrary-size-numeral` feature.

waywardmonkeys:
This is always on now.


waywardmonkeys:
@yozu @Pat-Lafon @wtdcode @TheVeryDarkness  Any objections to this?

PR #258, opened 2023-10-25
ci: Enable vcpkg build on macOS and Linux

waywardmonkeys:
cc: @TheVeryDarkness 

Seeing if we can enable this now...

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -86,11 +86,11 @@ jobs:
   build_with_vcpkg_installed_z3:
     strategy:
       matrix:
-        build: [macos, windows] # [linux, macos, windows]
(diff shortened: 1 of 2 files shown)


TheVeryDarkness:
Maybe it's a good start for distributing our built z3 static library, as I deliberately cache built z3 in CI.

TheVeryDarkness:
> Maybe it's a good start for distributing our built z3 static library, as I deliberately cache built z3 in CI.

And this fix make it possible to use our built z3 on Linux and macOS.

But it still remains a weird thing to me that on Linux and macOS, standard library won't be linked by default.

TheVeryDarkness:
And z3 version in vcpkg may change if we don't specify a concrete revision ID such as commit SHA.

PR #260, opened 2023-10-26
Rename `static-link-z3` to `bundled`.

waywardmonkeys:
This renames the `static-link-z3` build feature to `bundled` but leaves a bit around for now to maintain compatibility.

It also improves the README's discussion a little bit to mention `vcpkg`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -37,7 +37,7 @@ jobs:
       # XXX: Ubuntu's Z3 package seems to be missing some symbols, like
       # `Z3_mk_pbeq`, leading to linker errors. Just ignore this, I guess, until
       # we figure out how to work around it. At least we have the
-      # statically-linked Z3 tests below...
+      # build using a bundled Z3 below...
       if: ${{ success() || failure() }}
 
   build_on_wasm:
@@ -57,12 +57,12 @@ jobs:
         source ./emsdk_env.sh
     - name: Install wasm32-unknown-emscripten target
       run: rustup target add wasm32-unknown-emscripten
-    - name: Build z3-sys and z3 with statically linked Z3
+    - name: Build z3-sys and z3 with bundled Z3
       run: |
         source ~/emsdk/emsdk_env.sh
-        cargo build --target=wasm32-unknown-emscripten --features static-link-z3
+        cargo build --target=wasm32-unknown-emscripten --features bundled
 
-  build_z3_statically:
+  build_with_bundled_z3:
     strategy:
       matrix:
         build: [linux, macos, windows]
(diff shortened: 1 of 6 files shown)


waywardmonkeys:
FYI @TheVeryDarkness @hermanventer @Pat-Lafon @yasuo-ozu @wtdcode

This lays the ground for starting to use `pkg-config` to find the system-provided build (#248) and downloading pre-built Z3 binaries (#249).

TheVeryDarkness:
How about uploading the z3 built by vcpkg? It's statically linked. Then we can download it.

waywardmonkeys:
@TheVeryDarkness The Rust binary ABI isn't stable, so it would be specific to a particular version or set of versions of Rust. But getting #249 in place with downloading the Z3 binary might be good for what you want?

TheVeryDarkness:
> @TheVeryDarkness The Rust binary ABI isn't stable, so it would be specific to a particular version or set of versions of Rust. But getting #249 in place with downloading the Z3 binary might be good for what you want?

A question I've met with the binary downloaded from z3 release is that it seems to be dynamically linked. In rust, it's not recommended, and in my test cases, doctests might fail, as dynamic library may not be found, And distributing our program with a .dylib or a .dll may be a little annoying.

As far as I know, vcpkg is building z3 as a cpp project, therefore it's following cpp ABI. Maybe let me have a try. I'm already testing uploading and downloading our built z3.

Thanks for your answer.

waywardmonkeys:
Ah, just the `libz3.a` ... not the `z3-sys` or `z3` crate outputs ...

TheVeryDarkness:
> Ah, just the `libz3.a` ... not the `z3-sys` or `z3` crate outputs ...

Yep. Sorry, I didn't make it clear.

TheVeryDarkness:
Anyway, we can provide a lot of options.

Pat-Lafon:
Is it possible to add a compile-time warning from build.rs if it is built with the `static-link-z3` flag?

TheVeryDarkness:
> @TheVeryDarkness The Rust binary ABI isn't stable, so it would be specific to a particular version or set of versions of Rust. But getting #249 in place with downloading the Z3 binary might be good for what you want?

I think #249 is good enough if users are just using z3 for their own usage. Is there anything preventing it from being in place?

Days ago, I opened [an issue](https://github.com/Z3Prover/z3/issues/6897) in z3 repository, asking whether they can build a statically-linked one, but got a refusal.

TheVeryDarkness:
> Is it possible to add a compile-time warning from build.rs if it is built with the `static-link-z3` flag?

Maybe a `println!` or `eprintln!`?

waywardmonkeys:
Warning added:

```
$ cargo build --features static-link-z3
   Compiling z3-sys v0.8.1 (... z3-sys)
warning: The 'static-link-z3' feature is deprecated. Please use the 'bundled' feature.
   Compiling z3 v0.12.1 (... z3)
    Finished dev [unoptimized + debuginfo] target(s) in 1m 37s
```

PR #264, opened 2023-10-29
Upload z3 built by vcpkg as release assets in CI. Maybe test it in CI later.

TheVeryDarkness:
The release name `0.12.2` is in env.Z3_VERSION and may be changed to `z3-0.12.2` if expected.

Uploaded files in my fork is in [0.12.2](https://github.com/TheVeryDarkness/z3.rs/releases/tag/0.12.2).

I didn't attach a SHA256 or other validation information, and this may be added in the future.

And I may also add some codes to let z3-sys download and test uploaded archive. But archive for Windows is just too big, can we cache downloaded archive globally? Or add an env var that points to the archive？

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -99,11 +99,14 @@ jobs:
       matrix:
         build: [linux, macos, windows]
         include:
-          - os: ubuntu-latest
+          - build: linux
+            os: ubuntu-latest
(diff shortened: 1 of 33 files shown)


TheVeryDarkness:
The test failed because there is no such release 0.12.2 here and it don't have write access.

PR #271, opened 2023-11-10
Into i64

Pat-Lafon:
I wanted to concertize my idea from #270 into some code. Instead of creating additional `from_i*` methods I propose using `Into<i64>` for `from_i64`. Leveraging the trait system here runs into some non-obvious issues.

- One probably can't implement `from_u64` with `Into<u64>`. Well, it's allowed but then calling `Int::from_u64(1)` is a typing error because the type of `1` here is ambigious(under-constrained). When an integer literal has an ambigious type, it defaults to `i32` and therefore fails. This is fine for `Into<i64>` since `i32` implements `Into<i64>`
- It's rather non-trivial to try to write something like `impl<'ctx, T: Into<i64>> Div<T> for Int<'ctx> {` because there are other implementations of `Div` for `Int<'ctx>` that the compiler can't prove to not overlap(either currently or in some future version which would lead to breakage). Maybe there is something I'm missing or some kind of nightly negative trait impls that one could use. `ops.rs` already heavily makes use of macros so I figure it's find to make a new macro that generalizes over the valid integer types: `u64, i64, i32, i16, i8`

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -773,9 +773,9 @@ impl<'ctx> Int<'ctx> {
         }
     }
 
-    pub fn from_i64(ctx: &'ctx Context, i: i64) -> Int<'ctx> {
+    pub fn from_i64<T: Into<i64>>(ctx: &'ctx Context, i: T) -> Int<'ctx> {
         let sort = Sort::int(ctx);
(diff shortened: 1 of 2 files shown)


toolCHAINZ:
@Pat-Lafon is this still relevant with the new trait changes?

Pat-Lafon:
I think this has been subsumed by your latest work

PR #275, opened 2023-11-20
Support for more regular expression operations

### LINKED ISSUE #272: Support for regular expression "mk_re_all" and "mk_re_allchar"
Hi, I wonder is it possible to use the current RUST API to express the string "a*b?c", where "*" means any string, and "?" means any character? It seems to require the functions like "mk_re_all" and "mk_re_allchar", which are not supported by the current RUST API.

Pat-Lafon:
Fixes #272

(First two commits) My attempt at adding the new regular expression operations: `power`, `allchar`, and `diff`. I also noticed that `option` was exposed but not implemented so I've added that as well.

(Third commit) I was looking to test my inclusions to the code base but I realized that I'm not sure about the semantics of regular expressions in Z3? I ran into some weird behavior I don't understand so I just added a simple test from https://z3prover.github.io/api/html/z3py_8py_source.html#l11172 to get some advice. `test_regex_union` works as expected, checking that `(a|b)` matches with `a` and `b` but not `c`. However, I found that I get `SatResult::Sat` in more cases than I expect like in `test_regex_union2` which just checks that `c` matches with `(a|b)`. I would expect to get `SatResult::Unsat` but I don't?

### DIFF
--- a/z3-sys/src/lib.rs
+++ b/z3-sys/src/lib.rs
@@ -3124,7 +3124,7 @@ extern "C" {
     /// Create the regular language `re*`.
     pub fn Z3_mk_re_star(c: Z3_context, re: Z3_ast) -> Z3_ast;
 
-    /// Create the regular language `[re]`.
+    /// Create the regular language `re?`.
     pub fn Z3_mk_re_option(c: Z3_context, re: Z3_ast) -> Z3_ast;
 
     /// Create the union of the regular languages.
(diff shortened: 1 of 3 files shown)


waywardmonkeys (changes requested):
In upstream Z3, you have to look at some of the SMTLIB tests to find things that actually exercise this code (rather than via the native API).

waywardmonkeys:
@Pat-Lafon Do you want to try rebasing this on top of current `master` and see if CI passes? We're preparing for a new release.

PR #276, opened 2023-11-23
Add an environment variable `EMSCRIPTEN_ROOT` for emscripten installed with brew and add supports for other wasm targets.

TheVeryDarkness:
Homebrew only provides emscripten but not a full emsdk, so brew users may need an extra environment variable `EMSCRIPTEN_ROOT`.

In the past versions, emscripten is considered only when the target is wasm32-unknown-emscripten, but it may be needed for other wasm targets. I add supports for other wasm32 targets. And missing emscripten won't cause a panic for non *-emscripten targets.

And we can test it on wasm32 targets if vcpkg-rs supports them later. [Related PR](https://github.com/mcgoo/vcpkg-rs/pull/53) has been merged but has not been published to crates.io.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,6 +12,13 @@ env:
   RUSTFLAGS: "-D warnings"
 
 jobs:
+  check-formatting:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Check formatting
+      run: cargo fmt -- --check
+
   build:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -94,6 +94,46 @@ jobs:
         run: cargo test -vv --features static-link-z3
       - name: Test `z3` with statically linked Z3 and `arbitrary-size-numeral` enabled
         run: cargo test --manifest-path z3/Cargo.toml -vv --features 'static-link-z3 arbitrary-size-numeral'
+  build_with_vcpkg_installed_z3:
+    strategy:
+      matrix:
+        build: [linux, macos, windows]
+        config:
+          - os: ubuntu-latest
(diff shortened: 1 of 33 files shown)


waywardmonkeys (changes requested):
Would it be possible to rebase this against current `master` and eliminate a bunch of the commits so that I can see what a clean history might look like?

waywardmonkeys:
Since this is only used for getting a sysroot for running clang to generate the enum bindings, perhaps we can solve this in an even simpler way... Could we just copy a couple of files into an `z3-sys/ext/wasm-sysroot` to give a minimal `stdbool.h`, `stdint.h` and `stdio.h`?

Actually, I think I can submit a PR to upstream to remove in the include of `stdiio.h` from here...

TheVeryDarkness:
> Would it be possible to rebase this against current `master` and eliminate a bunch of the commits so that I can see what a clean history might look like?

Oh, I'm sorry. Let me take a glance at `git rebase`.

TheVeryDarkness:
> Since this is only used for getting a sysroot for running clang to generate the enum bindings, perhaps we can solve this in an even simpler way... Could we just copy a couple of files into an `z3-sys/ext/wasm-sysroot` to give a minimal `stdbool.h`, `stdint.h` and `stdio.h`?
> 
> Actually, I think I can submit a PR to upstream to remove in the include of `stdiio.h` from here...

Yep, I think those ways should be better.

waywardmonkeys:
I'm going to close this and will do the `wasm-sysroot` changes this weekend. Thanks for bringing this up!

(Also, I got a change merged upstream already to remove the inclusion of `stdio.h`)

TheVeryDarkness:
> I'm going to close this and will do the `wasm-sysroot` changes this weekend. Thanks for bringing this up!
> 
> (Also, I got a change merged upstream already to remove the inclusion of `stdio.h`)

Sounds great :)

PR #277, opened 2023-11-25
z3-sys: Use `pkg-config` when using system libs.

waywardmonkeys:
This helps find the lib and includes more accurately than not using it.

This builds on previous work by @poscat0x04 and @wtdcode.


waywardmonkeys:
This will replace the PR #248.

PR #280, opened 2023-12-18
sort: fix array_range and array_domain lifetimes

Philipp15b:
Fixes the lifetimes on `array_domain` and `array_range` functions on `Sort`. The lifetime should be `'ctx`, not the lifetime of `self`.


waywardmonkeys:
The test failure is due to some of the tests now using a more recent Z3 (partly because I pushed an updated version of Z3 into vcpkg). I'll submit my fix for that test failure in the next day or two. I'll review this separate from that.

waywardmonkeys (approved):

PR #281, opened 2023-12-18
add new_const and fresh_const functions to Dynamic

Philipp15b:
This is an alternative to #191, fixing #190. I simply added new functions to the existing `Dynamic` type.


waywardmonkeys:
If you want to submit a PR that adds more test coverage (or comments with doc tests), that would be great. But I'll merge this for now.

waywardmonkeys:
Also, thanks for this and the other PR!

PR #283, opened 2024-01-18
Optimize: Add parameter configuration API

Mrmaxmeier:
This allows configuring `Optimize` instances. I'm using this to enable the 'box' priority mode, but that'll require additional API changes.
(Other Z3 bindings return an "objective handle" and allow fetching lower and upper bounds. These handles are index-based though and are not reference counted, so I'm not sure about the best Rust API equivalent yet.)


waywardmonkeys (approved):

waywardmonkeys:
Thanks!

PR #287, opened 2024-03-29
Add creating reals from any numeral and getting f64 from reals

lucascool12:
Adds the ability to create Reals from any numeral using [Z3_mk_numeral](https://z3prover.github.io/api/html/group__capi.html#ga2adef1696d308843fdb33ad6b2d4b4a0).
Also adds the method `as_f64` to `Real` which uses [Z3_get_numberal_double](https://z3prover.github.io/api/html/group__capi.html#gaf48e7b61664c0273a4ec77649f0a00cd).

This is useful if you want to represent a real as a f64. Currently you would have to calculate the numerator and denominator in order to create a z3 real from a f64. With this merge request all you would need to do is create a string from your f64 and pass this to `from_str`.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -583,10 +583,17 @@ impl<'ctx> Real<'ctx> {
     }
 
     pub fn from_real_str(ctx: &'ctx Context, num: &str, den: &str) -> Option<Real<'ctx>> {
+        let fraction_string = format!("{num:} / {den:}");
+        Self::from_str(ctx, &fraction_string)
+    }
+
+    /// The string may be of the form \[num\]*\[.\[num\]*\]\[E\[+|-\]\[num\]...+\].
+    /// Or a rational, that is, a string of the form \[num\]* / \[num\]*
+    pub fn from_str(ctx: &'ctx Context, numeral: &str) -> Option<Real<'ctx>> {
         let sort = Sort::real(ctx);
         let ast = unsafe {
-            let fraction_cstring = CString::new(format!("{num:} / {den:}")).unwrap();
-            let numeral_ptr = Z3_mk_numeral(ctx.z3_ctx, fraction_cstring.as_ptr(), sort.z3_sort);
+            let c_numeral = CString::new(numeral).unwrap();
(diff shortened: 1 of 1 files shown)


z3/src/ast.rs
```diff
@@ -583,10 +583,17 @@ impl<'ctx> Real<'ctx> {
     }
 
     pub fn from_real_str(ctx: &'ctx Context, num: &str, den: &str) -> Option<Real<'ctx>> {
+        let fraction_string = format!("{num:} / {den:}");
+        Self::from_str(ctx, &fraction_string)
+    }
+
+    /// The string may be of the form \[num\]*\[.\[num\]*\]\[E\[+|-\]\[num\]+\].
```
waywardmonkeys, line 590: Could you put `...` around the `\[num\]...+\]` ?

waywardmonkeys (changes requested):
If you can make the suggested change and help get this to pass the relevant parts of CI, that'd be great.

lucascool12:
I fixed the relevant parts of the CI, and then accidentally clicked close pull request, whoops.

dragazo:
`Z3_get_numeral_double` fails (by returning zero) if the result cannot be stored perfectly in a 64-bit floating-point value. For example, the irrational number `sqrt(2)` would have this issue. If that's the intended behavior in this PR, we might want to call it something like `as_exact_f64` or `exact_f64`. As a user, I'd also want it to return `Option<f64>` instead of `f64` with a default `0.0` for failure - although I don't know if Z3 exposes a way to check if that operation succeeds...

But if that's not the intended behavior, I have a separate PR that works even for irrational numbers: #304.

lucascool12:
I didn't know `Z3_get_numeral_double` could fail. #304 is pretty much what I need for my use case.
Except that it is missing the ability to create a Real from a numeral string, I'll put this in a different PR.
Still we could keep this PR alive for exact f64 as you said.
The documentation specifies that the precondition of `Z3_get_numeral_double` is `Z3_get_ast_kind(c, a) == Z3_NUMERAL_AST || Z3_is_algebraic_number(c, a) `, I tried this but got unexpected results.
The method @dragazo described does seem to work.
I'll change to title of the PR.

lucascool12:
> I'll change to title of the PR.
Scratch that I'll close this one and create 2 new ones.

PR #292, opened 2024-05-17
Expose Z3_get_version in the high-level interface

JakobR:
I like to use this to check that the right Z3 library got linked, so it would be good to have it accessible from the high-level interface.


JakobR:
The CI failures seem unrelates to this PR.

waywardmonkeys (approved):

waywardmonkeys:
Thanks! And yes, they are ...

PR #295, opened 2024-05-28
Update bundled z3 (fix windows debug builds)

toolCHAINZ:
This seems to fix debug windows builds on my machine


waywardmonkeys:
Okay, let's give this part a try!

Issue #294, opened 2024-05-21
New release


toolCHAINZ:
I noticed it's been a while since a new version of these bindings have been pushed to crates.io and there's some really good stuff missing from that version.

Any possibility of a release in the near future?

waywardmonkeys:
I'd like to ... I want to get the bundled Z3 updated, but that is still failing and I need to figure out the right fix. Just now in fixing something else, the WebAssembly build failed due to an error in Z3 and I've just submitted a PR for that. (That will require a new release of Z3...)

Would be good to do though!

toolCHAINZ:
Gotcha, thanks! Please let me know if there's any way I can help out; I've loved using this library and wouldn't mind contributing if there was something I could do. Out of curiousity, what is failing? I just pulled down `z3.rs` and bumped `z3-sys/z3` to `z3-4.13.0` and all the `z3-sys` and `z3` tests passed (ARM macOS).

toolCHAINZ:
Poking around at this a bit on my Windows 11 machine (apologies if you've already triaged all this):

* `cargo build --features bundled` succeeds
* `cargo test --features bundled` fails
* `cargo test --release --features bundled` succeeds

It looks like for some reason the debug build is emitting a linker flag to link to `msvcrt.lib`, which is the release version of the library. But there seems to be a default link to the debug version of the same library?

```
  = note: LINK : warning LNK4098: defaultlib 'MSVCRTD' conflicts with use of other libs; use /NODEFAULTLIB:library
```

waywardmonkeys:
I know that it is from doing debug builds on Windows, but not much beyond that. I haven't been booting my Windows machine often as of late ... I also know that it started when I increased the cmake version requirements in upstream Z3 (and therefore being subject to changes in the cmake policies).

toolCHAINZ:
Ok, I just tracked down some `cmake` stuff. Bisected `cmake` versions and it seems that `3.15` is the first version with broken debug tests.

Still looking through the changelog for that version, but the following item does pop out:


[...]

PR #298, opened 2024-05-29
Expose underlying Z3_context and Z3_sort

### LINKED ISSUE #290: No way to access the raw `Z3_sort` from `Sort<'ctx>` or `Z3_context` from `Context`
Most functionality around `Float` types is unbound. Adding bindings without editing or forking this library is impossible without the extreme hack of declaring copies of the structs and transmuting references between them to trick Rust into accessing the private fields `z3_sort` or `z3_ctx`.

I was attempting to invoke `z3_sys`'s `Z3_mk_fpa_to_fp_real`, but it requires raw access to the `Z3_sort` values, as well as a rounding mode, but functions like `z3_sys::Z3_mk_fpa_rne` require an instance of the raw `Z3_context`, which lack a high-level wrapper.

JakobR:
This would be useful to use functions from the raw bindings that aren't exposed in the high-level bindings yet. Since the raw bindings are `unsafe`, it should be fine to give access to the `Z3_context` and `Z3_sort`.

Closes #290.

### DIFF
--- a/z3/src/context.rs
+++ b/z3/src/context.rs
@@ -17,6 +17,10 @@ impl Context {
         }
     }
 
+    pub fn get_z3_context(&self) -> Z3_context {
+        self.z3_ctx
+    }
+
     /// Interrupt a solver performing a satisfiability test, a tactic processing a goal, or simplify functions.
     pub fn interrupt(&self) {
         self.handle().interrupt();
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
Thanks! Per #294, we will try to do a new release soon.

PR #300, opened 2024-06-25
Z3 Optimize: add `assert_and_track` and `get_unsat_core`

toolCHAINZ:
I noticed these are missing from `Optimize` and would have some use for them.

### DIFF
--- a/z3/src/optimize.rs
+++ b/z3/src/optimize.rs
@@ -1,6 +1,7 @@
 use std::convert::TryInto;
 use std::ffi::{CStr, CString};
 use std::fmt;
+use log::debug;
 
 use z3_sys::*;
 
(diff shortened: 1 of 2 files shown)


z3/src/optimize.rs
```diff
@@ -105,6 +126,43 @@ impl<'ctx> Optimize<'ctx> {
         unsafe { Z3_optimize_minimize(self.ctx.z3_ctx, self.z3_opt, ast.get_z3_ast()) };
     }
 
+    /// Return a subset of the assumptions provided to either the last
+    ///
+    /// * [`Optimize::check`] call.
```
waywardmonkeys, line 131: This seems like it needs editing?

waywardmonkeys (changes requested):
Looks good to me apart from the comment… would it be easy to add a test for this?

z3/src/optimize.rs
toolCHAINZ: Ooops, I'll fix that!

toolCHAINZ:
> Looks good to me apart from the comment… would it be easy to add a test for this?

Ok, I've added a test that exercises both `assert_and_track + check` and an individual `check`.

waywardmonkeys:
> Ok, I've added a test that exercises both `assert_and_track + check` and an individual `check`.

Great, thanks!

PR #304, opened 2024-07-25
Real approx functions

dragazo:
It's a bit difficult to work with `Real` (in `z3`, not `z3-sys`) since the only provided way to get a value out at the end is `Real::as_real(&self) -> Option<(i64, i64)>`, which really only works for rational numbers. In this PR I've added two new functions to help:

- `Real::approx(&self, precision: usize) -> ::std::string::String`, which gives a string approximation of the real value up to `precision` correct digits after the decimal (via `Z3_get_numeral_decimal_string`).
- `Real::approx_f64(&self) -> f64`, which is just a convenience wrapper for `self.approx(17).parse().unwrap()` to get a full-precision `f64` approximation of the real number.

I also noticed that `tests/objectives.rs`, `tests/ops.rs`, and `tests/semver_tests.rs` weren't actually having their tests executed since they weren't linked into the lib, so I declared them as (test) modules. It looks like some of those are failing, but that's not something that this PR caused.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -940,7 +940,7 @@ impl<'ctx> Real<'ctx> {
         }
         .to_str()
         .unwrap();
-        s.strip_suffix("?").unwrap_or(s).to_owned()
+        s.strip_suffix('?').unwrap_or(s).to_owned()
     }
     pub fn approx_f64(&self) -> f64 {
         self.approx(17).parse().unwrap() // 17 decimal digits needed to get full f64 precision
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
I'll fix that clippy error and then you can rebase.

waywardmonkeys:
Oops, never mind ... that's in your code!

dragazo:
Sure, I think I just fixed the last of the clippy issues.

waywardmonkeys:
Thanks!

PR #305, opened 2024-07-26
changed: rename `_eq` to `eq` and `*_real_*` to `*_rational_*`

dragazo:
_Note: this PR contains breaking changes, so if merged we should at least change the minor version number before next release_

When using this crate, I thought it was confusing that operators are used for symbolic arithmetic but doing the same for checking equality is non-symbolic (ast comparison). To actually check symbolic equality, you need to use `_eq(other)`, which looks weird compared to other symbolic inequality checks like `le(other)` or `gt(other)`. So in this PR I've renamed `_eq` to `eq`, the only cost of which is removing the `PartialEq` trait impl from the concrete `Ast` types. Instead, I've added a new method to `Ast` called `ast_eq(&self, other: &dyn Ast) -> bool` that does this check explicitly. I think this is less likely to be used accidentally than `==`, for which the user probably wants symbolic equality anyway.

Also, there were several functions for `Real` that take/return a fraction but use the term "real" (e.g., `from_real(i32, i32)`). I've renamed these to "rational".

Also, for the new `from_rational` function (replacing `from_real`), I changed the input types to `i64` instead of `i32`, but the deprecated wrapper still uses `i32`. This is to match the return type of `as_rational` (new name for `as_real` but otherwise unchanged).

For all of the function renames, I kept a shallow wrapper with the old name and just marked it as deprecated. So the renames aren't a breaking change, but removing the `PartialEq` impl is.


waywardmonkeys:
Thanks!

I think this sounds like an improvement!

If you could:

* Rebase on top of `master`
* Squash your commits into a single one
* Make the commit message for that commit be like your description of this PR.

That would be great!

dragazo:
Sure, I just merged and squashed it back down to one commit.

toolCHAINZ:
@dragazo would you be willing to merge `master` into this one more time? Sorry, there's going to be a lot of conflicts. If not, would you mind me taking a try at it? These changes look excellent to me.

dragazo:
@toolCHAINZ I started merging changes, but a lot of tiny things have changed in many places, so it'll take a while to sift through. It might actually be faster to start from scratch and reapply these changes in the few places that need them. I'll try that sometime next week, but I don't mind if you'd like to take a crack at it.

toolCHAINZ:
Yeah agreed the operative changes look pretty minimal, just reapplying the change and refixing all the tests feels like the way to go. We'll see who gets to it first haha

toolCHAINZ:
I took a crack at this and ended up making some changes. I think I found a happy medium for this `eq` stuff. So the key observation is that `Rust` allows for multiple methods with the same name to be attached to a type: one on the type itself, and others on traits. It will always prefer to use the type's own method instead of that of a trait. If multiple traits match a method call, it will require explicit disambiguation.

So, I made the following changes:

* `ast_eq`, `eq`, and all of the `_` variants of `eq` and `safe_eq` are taken out of `Ast` and implemented directly on each implementing type through the `impl_ast!` macro.
* We keep the `PartialEq` and `Eq` trait impls on `Ast`s, which uses `ast_eq` internally.

This basically means you keep `==`, which I _do_ think is useful for cases like checking if a variable is in a vector. But if you call `eq`, you get a `Bool`, as expected.

dragazo:
@toolCHAINZ Oh, the original reason I removed `==` altogether isn't because of ambiguous resolution issues, but because for a solver `s`, `s.assert(a == b)` correctly (but unfortunately) compiles since `bool: Into<Bool>`. However, in this instance the user probably didn't actually want to do an ast equality assertion (esp. cause that's just a constant) and probably wanted a symbolic equality like `_eq`. I've run into that a few times where I basically either get `assert(true)` which does nothing or `assert(false)` which breaks the entire solver. So I opted to make `==` not even valid syntax to prevent the possible misuse.

toolCHAINZ:
Ohh that's a good point! Maybe a fix for this would be to make assert take a Bool and not an Into<Bool>

dragazo:
That would go a long way, as long as there aren't any other useful `Bool::from` things we'd lose out on (none that I can think of). As far as I can tell, I think the only thing that could throw that off is if the user tries to fix the type error by adding `.into()` themselves, but at least that makes it have to be both explicit and visible at the call site. Personally, I still lean towards making the ast version explicit like `ast_eq` and just not having `==`, but either way would be a huge improvement if the errors I was running into are a common thing others experience (not certain, I may be biased from debugging trauma lol).

toolCHAINZ:
Yeah I'm reluctant to remove the impl entirely as I think it is very useful to be able to use eq for Ast equality. We should go ahead with changing the assert arg to a `Bool`. I as well can't think of any reason you'd ever want to not just pass a `Bool` in. Are there any other places where we're accepting Into where it doesn't _really_ make sense?

dragazo:
I think the `Into<Bool>` one is probably the only really problematic one (in certain scenarios like this). The other ones are all fine as far as I can think of, as long as we don't have conversions between types like float <-> int (but iirc we don't). Ultimately, I think only primitives could be a problem, and bool is kind of special in that regard, and mostly just for assert. Though I mostly just do discrete optimization, so maybe others could weigh in if there are similar issues with reals like for nonlinear opt.

Issue #302, opened 2024-07-17
Add consequences to solver


VadeveSi:
Hi,

I wanted to work towards adding support for Z3's [consequences](https://z3prover.github.io/api/html/classz3_1_1solver.html#a4322874cf842109514596806684cd4d7) to the Solver. I figure something like

```rust
pub fn get_consequence(self,
                       mut assumptions: &[ast::Bool<'ctx>],
                       mut variables: &[ast::Bool<'ctx>]
                       ) -> &[ast::Bool<'ctx>] {
...
}
```

should work as a high-level wrapper around [`z3_sys::Z3_solver_get_consequences`](https://docs.rs/z3-sys/latest/z3_sys/fn.Z3_solver_get_consequences.html), but this requires the data to be in a [`Z3_ast_vector`](https://docs.rs/z3-sys/latest/z3_sys/type.Z3_ast_vector.html) and I'm struggling on how to do that. Do you have any suggestions on how to do this? Am I overlooking something obvious?

Thanks.

toolCHAINZ:
It looks like this was added in #308, which has now been released. Closing this issue.

PR #308, opened 2024-08-07
Support consequences API (#302)

VadeveSi:
Add consequences, allowing you to derive implied propositions from a set of assumed propositions.

I'm not an expert rust user, feel free to point out any improvements possible.


waywardmonkeys:
Thanks!

PR #310, opened 2024-10-01
Expose sequence sort and AST

yagehu:
This commit exposes the sequence sort in the high level z3 binding and adds the `Z3_mk_seq_nth()` low level binding.

### DIFF
--- a/z3-sys/src/lib.rs
+++ b/z3-sys/src/lib.rs
@@ -3098,7 +3098,8 @@ extern "C" {
     /// Retrieve from `s` the unit sequence positioned at position `index`.
     pub fn Z3_mk_seq_at(c: Z3_context, s: Z3_ast, index: Z3_ast) -> Z3_ast;
 
-    /// Retrieve from s the element positioned at position index. The function is under-specified if the index is out of bounds.
+    /// Retrieve from `s` the element positioned at position `index`.
+    /// The function is under-specified if the index is out of bounds.
     pub fn Z3_mk_seq_nth(c: Z3_context, s: Z3_ast, index: Z3_ast) -> Z3_ast;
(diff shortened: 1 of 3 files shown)


z3-sys/src/lib.rs
```diff
@@ -3098,6 +3098,9 @@ extern "C" {
     /// Retrieve from `s` the unit sequence positioned at position `index`.
     pub fn Z3_mk_seq_at(c: Z3_context, s: Z3_ast, index: Z3_ast) -> Z3_ast;
 
+    /// Retrieve from s the element positioned at position index. The function is under-specified if the index is out of bounds.
```
waywardmonkeys, line 3101: How does this differ from `Z3_mk_seq_at`?

Also, some backticks should be used like in the other docstrings for `s` and `index`.

z3/src/ast.rs
```diff
@@ -1658,6 +1666,70 @@ impl<'ctx> Set<'ctx> {
...
+    pub fn at(&self, index: &Int<'ctx>) -> Self {
+        unsafe {
+            Self::wrap(
+                self.ctx,
+                Z3_mk_seq_at(self.ctx.z3_ctx, self.z3_ast, index.z3_ast),
+            )
+        }
+    }
+
+    /// Retrieve from s the element positioned at position index.
```
waywardmonkeys, line 1700: This and the previous need some backticks as well...

z3-sys/src/lib.rs
yagehu: Fixed the docstrings. Also added a sentence about the difference to the high level binding docstring.

waywardmonkeys (approved):

PR #314, opened 2024-10-08
Add bindings for sequence foldl

yagehu:
`foldl` (`fold_left`) can be combined with the recently added `lambda_const`.


yagehu:
It looks like this API is only available starting from z3 4.13.2.

toolCHAINZ:
If we want to get this merged in, we might look at extending the strategy in https://github.com/prove-rs/z3.rs/pull/444.

I would advocate for not bumping the min z3 version up to 4.13.2 by default (at least until ubuntu updates to a reasonable version), but we could gate it behind a feature to unblock CI like in that MR perhaps?

PR #320, opened 2024-11-04
feat: atmost and atleast

lucascool12:
Adds bindings for `atleast` and `atmost`. These functions are generic over `IntoIterator<Item = &'a Bool<'ctx>>` so that calling code can potentially reduce the amount of allocations since we always have to collect the input of `Bool` into an  `Z3_AST` array.

Some simple tests for both functions have also been added.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -760,6 +760,48 @@ impl<'ctx> Bool<'ctx> {
     }
 }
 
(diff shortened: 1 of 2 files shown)


waywardmonkeys (approved):
Will merge once the clippy issue is fixed! Thanks for the contribution!

PR #324, opened 2024-11-25
Adjust lifetimes on `ModelIter` to make them more permissive

sgpthomas:
Currently, iterating over a Model creates `FuncDecls` that reference the lifetime of the reference of the model that they come from, rather than the lifetime of context it contains.

This means that it is impossible to use those `FuncDecls` after the model is dropped, even when the context is still around. From looking at the code, this just seems like a bug, but let me know if this is intentional.

This PR adjusts the lifetimes of `ModelIter` to also explicitly include the lifetime of the model reference. This way it can return `FuncDecl` that properly reference the lifetime of the context.

### DIFF
--- a/z3/src/model.rs
+++ b/z3/src/model.rs
@@ -130,7 +130,7 @@ impl<'ctx> Model<'ctx> {
         }
     }
 
-    pub fn iter(&'ctx self) -> ModelIter<'ctx> {
+    pub fn iter<'a>(&'a self) -> ModelIter<'a, 'ctx> {
         self.into_iter()
     }
 }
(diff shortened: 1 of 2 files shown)


waywardmonkeys (approved):
I think this looks good.

It would be good to have a test for this to at least compile in a way that didn't before so that we make sure that this doesn't regress.

sgpthomas:
Added a test that only compiles with this change. Let me know if it looks good to you. Also, is the `wasm` build something that I broke? Or can I ignore that?

waywardmonkeys:
Looks great and will merge it shortly.

I don't think you broke the WebAssembly build. I'll try to take a look at it today or tomorrow.

Thanks for the PR!

Issue #328, opened 2024-12-02
WASM Build Failure


toolCHAINZ:
I did a little bit of triage on the recent WASM build failures. I don't have a fix, but the cause seems to be this:

https://github.com/rust-lang/cc-rs/pull/1284

Seems to have happened because:
* z3.rs has no committed lock file or direct dependence on a version of `cc`.
* `rust-bindgen`, which is what is actually using `cc` just targets the major version`1.0`.
* `cc` `1.2.0` changed the flags passed to wasm targets, adding in `-fno-exception`

So the CI is pulling in `cc 1.2.0`, which is using that flag and failing on the `z3` build.

toolCHAINZ:
Adding an explicit dependence on `cc` `1.1` in `z3_sys` as follows seems to fix it:

```toml
cc = "~1.1"
```

toolCHAINZ:
I've made #329 with this change.

PR #329, opened 2024-12-02
Fix CI

toolCHAINZ:
Three things in this MR:

* Fix the WASM build by pinning cc to 1.1 (see #328 for details).
* Fix the existing "too many arguments" lint failure with an `#[allow()]` annotation on the offending function.
* Fix some new lifetime elision clippy lints introduced in rust 1.83.

### DIFF
--- a/z3-sys/Cargo.toml
+++ b/z3-sys/Cargo.toml
@@ -16,7 +16,8 @@ homepage = "https://github.com/prove-rs/z3.rs"
 repository = "https://github.com/prove-rs/z3.rs.git"
 
 [build-dependencies]
-bindgen = { version = "0.69", default-features = false, features = ["runtime"] }
+bindgen = { version = "0.70", default-features = false, features = ["runtime"] }
+cc = "~1.1"
 cmake = { version = "0.1.49", optional = true }
 pkg-config = "0.3.27"
 vcpkg = { version = "0.2.15", optional = true }
--- a/z3-sys/Cargo.toml
+++ b/z3-sys/Cargo.toml
@@ -17,6 +17,8 @@ repository = "https://github.com/prove-rs/z3.rs.git"
 
 [build-dependencies]
 bindgen = { version = "0.70", default-features = false, features = ["runtime"] }
+# Temporarily pinning cc to 1.1 as 1.2 and later use -fno-exceptions, which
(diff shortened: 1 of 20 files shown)


z3-sys/Cargo.toml
```diff
@@ -17,6 +17,7 @@ repository = "https://github.com/prove-rs/z3.rs.git"
 
 [build-dependencies]
 bindgen = { version = "0.70", default-features = false, features = ["runtime"] }
+cc = "~1.1"
```
waywardmonkeys, line 20: Could you add a comment here about why this is pinned, so that we remember easier and can go back and fix it more correctly?

z3/src/ast.rs
```diff
@@ -2317,6 +2317,7 @@ pub fn exists_const<'ctx>(
 /// let f_f_3: ast::Int = f.apply(&[&f.apply(&[&ast::Int::from_u64(&ctx, 3)])]).try_into().unwrap();
 /// assert_eq!(3, model.eval(&f_f_3, true).unwrap().as_u64().unwrap());
 /// ```
+#[expect(clippy::too_many_arguments)]
```
waywardmonkeys, line 2320: This is basically setting an MSRV of 1.81 on the crate. I'm probably fine with that.

toolCHAINZ: This is a good point; I forgot this was introduced this recently. I can change to `allow` if you prefer.

waywardmonkeys: I am fine with it as is and we see if anyone comments. There isn’t much point for a crate like this one to worry about older compilers.

z3-sys/Cargo.toml
toolCHAINZ: Added a note in [531708c](https://github.com/prove-rs/z3.rs/pull/329/commits/531708c4d37f1dc0623f95e42eafb7b73b71ec32)

z3/src/ast.rs
toolCHAINZ: I went ahead and reverted this to `allow` as the package itself isn't currently specifying an MSRV and I don't want to break anyone's workflow without reason.

waywardmonkeys (approved):
Looks good, thanks!

PR #333, opened 2025-01-05
Add missing API for BV

### LINKED ISSUE #332: Missing API for BitVectors
Z3 can make bitvectors from arrays of bools. Something like this should work (I haven't tested it yet),

I am not sure about endianness, and how it will work
```rust
pub fn from_bit_vec(ctx: &'ctx Context, bit_vec: Vec<bool>) -> Option<BV<'ctx>> {
    let ast = unsafe {
        let size = bit_vec.len() as u32;
        let bits = bit_vec.as_ptr();

        let numeral_ptr = Z3_mk_bv_numeral(ctx.z3_ctx, size, bits);
        if numeral_ptr.is_null() {
            return None;
        }

        numeral_ptr
    };
    Some(unsafe { Self::wrap(ctx, ast) })
}

pub fn from_byte_vec(ctx: &'ctx Context, byte_vec: Vec<u8>) -> Option<BV<'ctx>> {
    let mut bit_vec = Vec::with_capacity(byte_vec.len() * 8);
    for item in byte_vec {
        for offset in 0..8 {
            let mask = 1 << offset;
            let bit = item & mask != 0;
            bit_vec.push(bit);
        }
    }

    Self::from_bit_vec(ctx, bit_vec)
}
```

I will make a PR when I test it.

the-ssd:
Added missing API for creating bitvectors from vector of bits and bytes

Fixes #332

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -1346,6 +1346,34 @@ impl<'ctx> BV<'ctx> {
         Some(unsafe { Self::wrap(ctx, ast) })
     }
 
+    pub fn from_bit_vec(ctx: &'ctx Context, bit_vec: Vec<bool>) -> Option<BV<'ctx>> {
+        let ast = unsafe {
+            let size = bit_vec.len() as u32;
+            let bits = bit_vec.as_ptr();
+
+            let numeral_ptr = Z3_mk_bv_numeral(ctx.z3_ctx, size, bits);
+            if numeral_ptr.is_null() {
+                return None;
+            }
+
+            numeral_ptr
+        };
+        Some(unsafe { Self::wrap(ctx, ast) })
+    }
+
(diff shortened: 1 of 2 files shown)


the-ssd:
I changed the order of processing for byte_vec because that is how it works internally in rust.

This is how 100 represented in [Allocation](https://doc.rust-lang.org/nightly/nightly-rustc/stable_mir/ty/struct.Allocation.html)
```
bytes = [
    100,
    0,
    0,
    0,
]
```

toolCHAINZ:
Hello! Thanks for the contribution (and sorry for the delay)!

We ended up merging an equivalent definition of the `from_bit_vec` API as `BV::from_bits` in #398 and #399.

There's not yet a version of the second API you provided, but I think I would want such an API to be explicit about the endianness (much like how `u64` explicitly has `from_le_bytes` and `from_be_bytes`)

PR #337, opened 2025-02-11
Update DeclKind in z3-sys for new values

twizmwazin:
This adds all of the existing DeclKinds defined in the C API to the DeclKind enum. I did not add comments as they seemed inconsistent in the original. Let me know if I can improve this, I intend to add higher level bindings to various ops here for my use case.

Future work: This feels like it should be done as a macro.

### DIFF
--- a/z3-sys/src/lib.rs
+++ b/z3-sys/src/lib.rs
@@ -544,6 +544,8 @@ pub enum DeclKind {
     /// produces an index, such that the arrays
     /// are different if they are different on the index.
     ARRAY_EXT = generated::Z3_decl_kind::Z3_OP_ARRAY_EXT as u32,
+    SET_HAS_SIZE = generated::Z3_decl_kind::Z3_OP_SET_HAS_SIZE as u32,
+    SET_CARD = generated::Z3_decl_kind::Z3_OP_SET_CARD as u32,
     /// Bit-vector numeral.
     BNUM = generated::Z3_decl_kind::Z3_OP_BNUM as u32,
     /// One bit bit-vector.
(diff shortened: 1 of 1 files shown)


twizmwazin:
I believe the ubuntu CI failure is due to Ubuntu 24.04 packaging z3 4.8.12, released in 2021, before some of these DeclKinds were added... I would advocate for bumping the minimum to something more recent. I don't know if that is palatable or conflicts with other goals though.

toolCHAINZ:
I ran into this same z3 issue on a project of mine. In my case, I re-targeted my docker setup on debian `trixie`. That distro is still unstable, but has an up-to-date z3 and has worked well (so far 🤞) for testing software using `z3.rs`.

twizmwazin:
In my case I will be statically linking z3, so I'm basically always using the bundled option.

toolCHAINZ:
I think we'll pull the trigger on this in April when the next LTS updates their ancient apt z3 package.

toolCHAINZ:
Thanks for contributing this @twizmwazin ! I'm going to be closing it unmerged because I've made some other changes that encompass this.

As of #510, we have both bindgen enums and functions checked in, and will be tracking these changes in upstream z3 more closely (now asserting a minimum supported version of 4.16.0). I've also added a `z3-src` crate wrapping the cmake logic for vendored builds, which `z3-sys` now uses as well. I've not yet published the new version to crates.io, but will soon.

I've seen you have a rust claripy equivalent you're writing; `z3-src` and/or `z3-sys` might now be a suitable backing for your crate.

twizmwazin:
Great to hear, thank you!

PR #340, opened 2025-03-27
Add Order instantiation to FuncDecl

grahnen:
FuncDecl for partial/linear/piecewise linear/tree orders.

Z3 specializes these, this is better than adding assertions manually :)

Please help me double check that this does what it's supposed to; and that it does it in a good way.

I am not well versed with the internals of z3 or with this library.

[API Reference](https://z3prover.github.io/api/html/z3__api_8h.html), section "Special relations"

### DIFF
--- a/z3-sys/src/lib.rs
+++ b/z3-sys/src/lib.rs
@@ -2255,6 +2255,22 @@ extern "C" {
         body: Z3_ast,
     );
 
+
+    /// Create a FuncDecl representing a partial order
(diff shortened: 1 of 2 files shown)


toolCHAINZ:
@grahnen thanks for the PR! Would you be willing to write some unit tests for these new APIs? This will also be useful as an example for how they should be used.

toolCHAINZ:
I went ahead and added a doc test for partial_order and some doc comments to each API pointing to the others in this family, as well as pointing to the z3 special relations page as a reference.

Thanks for contributing this!

PR #344, opened 2025-04-08
Add bindings to the user propagator

### LINKED ISSUE #343: Retrieve non-unsat proof attempts via on_clause callback
I'm working on analyzing solver behavior in cases where the result is `sat` or `unknown`, and I'd like to retrieve some form of proof or trace to understand what went wrong or what path the solver took. However, `Solver::get_proof()` is only available when the result is `unsat`, which limits post-analysis in other cases.

To address this, I was exploring the clause inference callback mechanism, as described in the [Z3 documentation](https://microsoft.github.io/z3guide/programming/Proof%20Logs/) (via `Z3_solver_register_on_clause` in the C API). This seems like a viable way to observe proof steps.

However, I couldn't find any support for this functionality in `z3.rs`.
Is there a recommended way to hook into clause inferences from Rust using this crate, or would I need to manually bind and call the `Z3_solver_register_on_clause` C function myself?

I'd appreciate any guidance and would be happy to help contribute an implementation if this sounds like a worthwhile addition.

Thanks!

puyral:
This adds bindings and an API for the user propagator and the `on_clause` callback.

I would consider this still experimental, as I'm not fully sure of what `z3` might do with my objects (e.g., weird lifetimes and thread safety).

This also closes #343

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,9 @@ target
 Cargo.lock
 *~
 .z3-trace
+.DS_Store
+
+# nix
+.envrc
+.direnv
+result
--- a/z3-sys/src/lib.rs
+++ b/z3-sys/src/lib.rs
@@ -1569,7 +1569,7 @@ pub type Z3_pop_eh = ::std::option::Option<
     unsafe extern "C" fn(
         ctx: *mut ::std::ffi::c_void,
         cb: Z3_solver_callback,
-        num_scopes: ::std::ffi::c_uint,
+        num_scopes: ::std::os::raw::c_uint,
     ),
 >;
 pub type Z3_fresh_eh = ::std::option::Option<
@@ -1605,7 +1605,7 @@ pub type Z3_decide_eh = ::std::option::Option<
         cyx: *mut ::std::ffi::c_void,
         cd: Z3_solver_callback,
         t: Z3_ast,
-        idx: ::std::ffi::c_uint,
+        idx: ::std::os::raw::c_uint,
         phase: bool,
     ),
 >;
@@ -1613,8 +1613,8 @@ pub type Z3_on_clause_eh = ::std::option::Option<
     unsafe extern "C" fn(
         ctx: *mut ::std::ffi::c_void,
         proof_hint: Z3_ast,
-        n: ::std::ffi::c_uint,
-        deps: *const ::std::ffi::c_uint,
+        n: ::std::os::raw::c_uint,
+        deps: *const ::std::os::raw::c_uint,
         literals: Z3_ast_vector,
     ),
 >;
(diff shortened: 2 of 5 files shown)


puyral:
So apparently Ubuntu ships with a version of `z3` that is too old for the user propagator...

puyral:
So I discovered that we can get null pointer back from z3 with my implementation of `fresh_eh` (which will result in a crash).

TL;DR: current lifetimes don't match `z3`'s and I don't yet have a solution to make them agree with Rust's borrow checker assumptions. 

Currently, it actually returns the null pointer that z3 gives back. This was done on purpose because implementing this function causes me huge lifetimes problems.

To give an idea of the problem. The C++ implementation for `fresh` says:

> `user_propagators` created using `fresh()` are created during search, and their lifetimes are restricted to search time. They should be garbage collected by the propagator used to invoke `fresh()`. The life-time of the `Z3_context` object can only be assumed valid during callbacks, such as `fixed()`, which contains expressions based on the context.

In practice, z3 gives a new `Z3_context` that the inner solver will use. I understand that this context may get garbage collected before we reach the end of the search. So, no `z3` object should "escape" the user propagator.
To add to the problem, it's our responsibility to garbage collect the subsolver's user propagator. We don't know when we should drop it, aside from the fact that it should happen before the very end of the search. At that point, that user propagator is almost sure to hold invalid pointers/references (e.g., the `Context`) that might cause UB during `Drop`.

These problems tend to cascade effects to the rest of the user propagator API.

I'm switching this into a draft since I don't have an immediate solution.

If you are using `z3` in a way that won't spawn subsolvers, then you are safe to use the current API.

toolCHAINZ:
It may be worth revisiting this in light of #401, #404, and #417.

Now that contexts are refcounted and implicit, there shouldn't be an issue dealing with arcane lifetimes. The trick would be designing the API to ensure correctness and memory safety in the face of callbacks that use Contexts we do not control.

I haven't done a detailed read of the contribution you have here, but going from your description of the difficulty, there's a general approach that I think could work:

* Change `ContextInternal` to an `enum`, with an internal "owned" variant (what it has right now) and a "borrowed" variant (which can be _unsafely_ created from a raw `Z3_Context`). The "owned" variant would behave as it does now, destroying the inner context on `Drop`. The "borrowed" variant would not, making it suitable for cases like this where we are given a `Context` that Z3 itself is managing.
* Any Rust code running in this user propagator context would have to be expressed in a `Send + Sync` closure. There would be a wrapper that receives the Z3-managed `Context`, builds a "borrowed" `Context` out of it, and runs the closure using that `Context` in `Context::thread_local`.

This would ensure that no `Ast` or `Solver` associated with that `Context` can escape, and if existing stuff outside the propagator is needed for these callbacks, they can be moved into the closure with `synchronized()/recover()`.

puyral:
Amazing! It sounds like this does indeed solve most of my issues!

I'll have a try at this Weekend

puyral:
I have adapted the work to the new RC-based `Context`. I also added a wrapper around the `UserPropagator`s to be able to garbage collect them.

All looks good, but my example segfaults somewhere in `z3` which doesn't seem to be related to the user propagator...

PR #352, opened 2025-07-04
feat: add gh-release feature to get z3 from released libraries

lmondada:
I am suggesting adding a new way of obtaining a copy of `z3`, using the pre-compiled libraries available from [z3's GitHub releases](https://github.com/Z3Prover/z3/releases).

This is IMO the "simplest" way to use `z3-sys` for supported platforms:
- no system-wide install of `z3` nor fiddling with environment variables required
- no compilation of the whole of `z3` (and dependency on C++ tooling that goes along with that) required
- one-off small cost of library download (~30-40MB), as well as one-off build costs of `build.rs` script

Further remarks:
- I've added a `Z3_SYS_Z3_VERSION` env variable to specify which z3 release to pull the libraries from (only works for `gh-release` feature)
- I've deactivated the generation of `cargo-rerun-if` instructions by bindgen in the case of `gh-release`. They are unnecessary (assuming GitHub released files aren't modified, detecting changes on `build.rs` and `Z3_SYS_Z3_VERSION` is sufficient) and were causing constant rebuilds on my machine, not sure why.
- I've made the features "bundled", "vcpkg" and "gh-release" mutually exclusive by checking that at most one of them is set in `build.rs`. Let me know if you'd rather I remove this.

### DIFF
--- a/z3-sys/build.rs
+++ b/z3-sys/build.rs
@@ -79,6 +79,8 @@ fn link_against_cxx_stdlib() {
 
 #[cfg(feature = "gh-release")]
 mod gh_release {
+    use std::path::Path;
+
     use reqwest::blocking::{Client, ClientBuilder};
 
     use super::*;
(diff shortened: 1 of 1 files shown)


lmondada:
The clippy failure is unrelated to this PR (I think a recent clippy update/buff)

toolCHAINZ:
Not a maintainer (but a contributor) and I think this looks great! In my past contributions when I've run into clippy issues (from CI `clippy` version bumps), I've just fixed them in the same PR and that's been fine. So you could probably do that here!

One question/comment on the code: now that there's several different flags effecting the definition of `header` in `build.rs`, I feel it's getting a little difficult to follow/reason about. Maybe this could be defined in a single place with an `if cfg!` chain instead (or maybe there's some way to shove this info into an enum so we could `match` on it)?

waywardmonkeys:
I prefer a separate standal9ne clippy fix and will do that today

waywardmonkeys:
Clippy lints that are failing now fixed in #353. Fixing a new compiler warning that is not yet in stable in another PR.

lmondada:
@toolCHAINZ Thank you for the suggestion, I've cleaned up `build.rs`. Agreed it looks much better now!

@waywardmonkeys I've merged in the clippy fix, thanks!

I've also made a small change that tries to reuse the downloaded files if they are left over from a previous build.

lmondada:
ah, looks like CI is getting rate limited by GitHub :/

toolCHAINZ:
Reviewing this code, this seems very reasonable to me. I like that the releases are downloaded into their own folders in `OUT_DIR`, very clean. I think this also makes another existing PR obsolete: #249

I have no problem merging this unless @waywardmonkeys has some objection.

Issue #345, opened 2025-04-13
`NotExist("/opt/homebrew/Cellar/z3/4.13.0/include/z3.h")` error on macos


cdstanford:
Apologies if I've missed an easy solution. On MacOS (15.3.2 Sequoia on M3) I get the following error when adding this to a Rust project:
```
$ cargo build
    Blocking waiting for file lock on build directory
   Compiling z3-sys v0.8.1
error: failed to run custom build command for `z3-sys v0.8.1`

Caused by:
  process didn't exit successfully: `/Users/caleb/<my_project>/target/debug/build/z3-sys-2589da465cd26a1c/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=build.rs
  cargo:rerun-if-env-changed=Z3_SYS_Z3_HEADER
  cargo:rerun-if-changed=/opt/homebrew/Cellar/z3/4.13.0/include/z3.h
  cargo:rerun-if-env-changed=TARGET
  cargo:rerun-if-env-changed=BINDGEN_EXTRA_CLANG_ARGS_aarch64-apple-darwin
  cargo:rerun-if-env-changed=BINDGEN_EXTRA_CLANG_ARGS_aarch64_apple_darwin
  cargo:rerun-if-env-changed=BINDGEN_EXTRA_CLANG_ARGS

  --- stderr
  thread 'main' panicked at /Users/caleb/.cargo/registry/src/index.crates.io-6f17d22bba15001f/z3-sys-0.8.1/build.rs:47:14:
  Unable to generate bindings: NotExist("/opt/homebrew/Cellar/z3/4.13.0/include/z3.h")
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This suggests that the build file cannot find my Z3 installation, but the options [here](https://github.com/prove-rs/z3.rs/tree/master/z3) didn't seem to work for me. I have done `brew install z3` and the installation suggests adding the `bundled` or `vcpkg` features, but when I do this I get:
```
the package <name> depends on `z3`, with features: `bundled` but `z3` does not have these features.

failed to select a version for `z3` which could resolve this conflict
```

Any idea what might be causing this error? 

Thanks!
Caleb

cdstanford:
I think I just need to point it to the latest z3?

```
ls /opt/homebrew/Cellar/z3/       
4.14.1
```

cdstanford:
Update: I was able to fix this for my setup as follows:

[...]

PR #364, opened 2025-07-14
docs: Better documentation of z3 installation options

lmondada:
Hi there!

I've noticed a typo in the main `z3` readme concerning installation options (it said there were 3 instead of 4 installation options). I've fixed that and taken the opportunity to expand the section a bit.

Thanks a lot for the recent flurry of work, it's great to see life in this project!


lmondada:
I think this would also close #345

Issue #348, opened 2025-06-05
z3.h not found on Ubuntu 25


mossbiscuits:
It works great on Ubuntu 24.04, but when I upgraded to 25.04, I saw the error `z3.h not found` during `cargo build`. This might be an issue with the z3 versions:
- Ubuntu 24 apt: 4.8.12 (works)
- Ubuntu 25 apt: 4.13.3 (broken)
- Latest on Github: 4.15.1 (works)

## Minimal Working Example of cargo.toml
```toml
[package]
name = "project"
version = "0.1.0"
edition = "2024"

[dependencies]
z3 = "0.12.1"
```

I installed z3, clang, and libclang-dev from apt, and `cargo build` fails with linker errors due to `z3.h`.

## Workaround
Running `cargo build` is successful when building z3 from scratch using the instructions at https://github.com/Z3Prover/z3?tab=readme-ov-file#building-z3-using-make-and-gccclang (Building Z3 using make and GCC/Clang). The sequence of commands that is successful for me on a fresh install of Ubuntu 25 to add z3 to a fresh Rust project is:
```bash
git clone https://github.com/Z3Prover/z3.git
cd z3
python scripts/mk_make.py
cd build
make
sudo make install
cargo new project
cd project
cargo add z3
cargo build
```

toolCHAINZ:
Just to double-check how have you installed `z3` in the ubuntu 25 install? I just tried in an ubuntu 25 docker container and was able to get it to work.

I _did_ run into the header message you mention, but that was because I had erroneously installed the `z3` apt package (the command line utility) instead of `libz3-dev`.

mossbiscuits:
I only had `z3` installed. I completely forgot about `libz3-dev`. That works for me as well. Thank you for your response!

PR #376, opened 2025-07-16
Add feature to use manually-built Z3 library

Evian-Zhang:
The `LIBRARY_PATH` environment variable mentioned in README has no effect. In fact, we could provide a manual feature for users to provide their own Z3 library for us to link to.

### DIFF
--- a/z3-sys/Cargo.toml
+++ b/z3-sys/Cargo.toml
@@ -28,6 +28,7 @@ vcpkg = { version = "0.2.15", optional = true }
 bundled = ["dep:cmake"] # Build Z3 via our bundled submodule.
 vcpkg = ["dep:vcpkg"] # Build Z3 via vcpkg.
 gh-release = ["dep:reqwest", "dep:serde_json", "dep:zip-extract"] # Download pre-compiled Z3 lib from GitHub release for supported platforms.
+manual = [] # Use Z3 built in advance manually
 
 # Legacy feature for short term compatibility
 static-link-z3 = ["bundled", "deprecated-static-link-z3"]
--- a/z3-sys/README.md
+++ b/z3-sys/README.md
@@ -27,7 +27,7 @@ $ cargo add z3-sys
 
 **Note:** This library has a dependency on Z3.
 
-There are 4 ways for this crate to currently find Z3:
+There are 5 ways for this crate to currently find Z3:
 
 * By default, it will look for a system-installed copy of Z3.
   On Linux, this would be via the package manager. On macOS, this
@@ -51,6 +51,9 @@ There are 4 ways for this crate to currently find Z3:
     provide it to the `READ_ONLY_GITHUB_TOKEN` environment variable. The
     `build.rs` step will automatically use this token (if present) to prevent
     throttling.
+* Enabling the `manual` feature requires user provide a pre-built Z3.
+  * Users should specify both `Z3_SYS_Z3_INCLUDE` and `Z3_SYS_Z3_LIB` environment
+  variables indicating the header search path and library search path to find Z3.
 
 **Note:** This crate requires a `z3.h` during build time.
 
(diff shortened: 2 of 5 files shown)


toolCHAINZ:
I'm not sure if I understand the intent here; the default (non-`vcpkg`/`gh-release`/`bundled`) build _does_ use a pre-built Z3 and adds it as a dynamic object dependency. This is actually how I use the library almost exclusively, on x86 and arm, linux and macos, so I feel pretty confident in saying this is working. If it's not, it's possible you don't have a development-friendly z3 installed (see #348) or that your environment isn't set up to allow bindgen to find the headers or the linker to find z3 (hence the notes about the library path). If you have one in a custom location, those same environment variables should let you point the build at it.

 The default build assumes:

* that the z3 headers are in the include path
* that the z3 shared object is in the library path

The LIBRARY_PATH note in the readme is specific to linux, we should probably add a note about that.

Evian-Zhang:
Thank you for you quick response!

I do not use any package manager to build z3 since I want to keep my global environment clean. As a result, I just build and installed z3 in custom location, and want to use this crate to link to it.

I'm afraid that Cargo does not use LIBRARY_PATH as expected, at least in my own settings (Ubuntu 22.04) it does not work. I also do not see any documentation that Cargo would follow such environment to link libraries.

toolCHAINZ:
Out of curiosity did you set the paths directly in your environment, or did you put them in `.cargo/config.toml`? I've had broken macos environments before where I had to do this myself (by setting `DYLD_FALLBACK_LIBRARY_PATH` in that same file), which fixed it.

Evian-Zhang:
I set it in .cargo/config.toml

toolCHAINZ:
Ok, I ran through an example in docker. Slightly different linux version, but still debian-based so should be roughly the same as your setup. I'm on an arm mac so there's an arm-specific path, but I think this shows the intent still:

```Dockerfile
FROM rust:bookworm

# Install build deps, Z3
RUN apt-get update && apt-get install -y \
    z3 \
    libz3-dev \
    build-essential \
    curl \
    clang \
    ca-certificates

# Move Z3 headers/libsto /opt/custom/z3
RUN mkdir /opt/custom
RUN mkdir /opt/custom/z3
RUN mkdir /opt/custom/z3/include
RUN mkdir /opt/custom/z3/lib
RUN cp /usr/include/z3* /opt/custom/z3/include
RUN cp /usr/lib/aarch64-linux-gnu/libz3.so.4 /opt/custom/z3/lib/libz3.so

# Remove Z3
RUN apt-get remove -y libz3-dev
```

In this Dockerfile, cargo cannot find the header or library when building a crate using `z3.rs` as a dependency, as you describe.

However, if you put the following in `.cargo/config.toml` (inside the crate's directory or the user's home folder) it will build, link, and run successfully:

```toml
[env]
Z3_SYS_Z3_HEADER="/opt/custom/z3/include/z3.h"
LIBRARY_PATH="/opt/custom/z3/lib"
```

Evian-Zhang:
Strange... I will dig it tomorrow. Thank you anyway!

Evian-Zhang:
I know the reason. I have a globally installed z3 which is a dependent of something else. The linker would seek to that in prior to the path I specified in LIBRARY_PATH. What if I want to use my locally built z3 at specific location? If we use cargo:rustc-link-search in build.rs, this would solve the problem since it takes priority in linking.

toolCHAINZ:
Ahhh interesting. I'll read up on that. My inclination is to add this as an optional environment variable in the default build (e.g. `Z3_LIBRARY_PATH_OVERRIDE`) and set the cargo flag if the env variable is set rather than add another crate feature. I think that would work for your use case from what you describe?

Evian-Zhang:
Yes, I think that would solve the problem :) Moreover, with that env, we may not need LIBRARY_PATH any more?

Evian-Zhang:
See #377.

PR #381, opened 2025-07-17
Bump to Rust 2024 edition

Evian-Zhang:
See [this discussion](https://github.com/prove-rs/z3.rs/discussions/372#discussioncomment-13784382).

I was going to split the bump into two steps as @Pat-Lafon suggested, but I found out that bumping from 2018 to 2021 edition is so smooth that the only thing we need to do is add `resolver = 2` in the workspace `Cargo.toml`. So I think we can just make it one PR.

At least, now we have better fmt :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,3 @@
 [workspace]
 members = ["z3", "z3-sys"]
+resolver = "2"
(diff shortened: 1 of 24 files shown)


toolCHAINZ:
I went ahead and added `rust-version` flags and updated the resolver to 3, which is the new default in edition 2024. We might as well use it.

I did a little rust version bisect search and previous to this MR, the MSRV seems to have been 1.71. After, it's 1.85 due to the edition requirement (and we aren't using anything added subsequently to that).

Evian-Zhang:
Great! Thank you!

PR #387, opened 2025-07-25
Add i128 and u128 Support for Integer ASTs

mehrad31415:
## Add i128 and u128 Support for Integer ASTs

This PR extends the `Int<'ctx>` API to support 128-bit integers.

- `from_i128(ctx: &Context, v: i128) -> Int<'ctx>`  
  Creates a Z3 integer numeral from a signed 128-bit integer.

- `from_u128(ctx: &Context, v: u128) -> Int<'ctx>`  
  Creates a Z3 integer numeral from an unsigned 128-bit integer.

- `as_i128(&self) -> Option<i128>`  
  Attempts to parse the internal numeral into a signed `i128`. Returns `None` if the AST is not a numeral or parsing fails.

- `as_u128(&self) -> Option<u128>`  
  Same as above, but for unsigned `u128`.


toolCHAINZ:
I'll look at merging this in post-#410 because there's some overlap there and I don't want to fix the merge twice.

PR #388, opened 2025-07-25
Add Int–Real Mixed Arithmetic Operations

mehrad31415:
## Add Mixed Int–Real Arithmetic Operations with Operator Overloading

Implemented mixed `Int`–`Real` overloads for the following operations:
- Addition (`+`)
- Subtraction (`-`)
- Multiplication (`*`)
- Division (`/`)
- Exponentiation (`pow`)

For each operation, both `Int op Real` and `Real op Int` are supported, with all ownership combinations (by reference and by value). The result is always a `Real<'ctx>` node.

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -1,12 +1,14 @@
 //! Abstract syntax tree (AST).
 
 use log::debug;
+use num::pow::Pow;
 use std::borrow::Borrow;
 use std::cmp::{Eq, PartialEq};
 use std::convert::{TryFrom, TryInto};
 use std::ffi::{CStr, CString};
 use std::fmt;
 use std::hash::{Hash, Hasher};
+use std::ops::{Add, Div, Mul, Sub};
 
 pub use z3_sys::AstKind;
 use z3_sys::*;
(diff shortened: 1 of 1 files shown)


toolCHAINZ:
I'll look over this, but I will likely want to go another way with these impls to follow the pattern introduced in #410. I think this could all be done fairly simply with some `Add<IntoAst<Real>>` style impls in that PR. After that's merged, I'll look into this, but might end up changing it a good bit/doing it in its own branch

toolCHAINZ:
I read into some background here and, while I think we should support this, I think I'd like to go another way.

Looking at the SMTLIB standard [here](https://smt-lib.org/theories-Reals_Ints.shtml), the mixed int/real theories don't actually mix Sorts in operations. They just provide `Real => Int` and `Int => Real` conversions that can then be used to produce the proper `Sort` for each pure-int or pure-real operation.

Looking at the Z3 source code and stack-overflow/reddit/etc it seems that, as an ergonomic feature, Z3 allows for giving `Asts` of mixed integer/real sorts on _some_ operations, but it is just inserting an `Int => Real` conversion under the hood.

Therefore, I think the way to go for this is to use that int-to-real low API to implement `Into<Real> for Int`. Then we don't need any extra impls and it should "just work". The generated `Asts` will also have explicit conversions indicating what sort stuff is, which I personally prefer to the implicit style.

PR #392, opened 2025-07-25
Add Constructors for Floating-Point NaN Values

mehrad31415:
- `Float::nan(ctx: &Context, sort: &Sort) -> Float<'ctx>`  
  Constructs a NaN value for any floating-point sort. 

- `Float::nan32(ctx: &Context) -> Float<'ctx>`  
  Convenience method for producing a NaN in IEEE-754 single precision (8 exponent bits, 24 significand bits).

- `Float::nan64(ctx: &Context) -> Float<'ctx>`  
  Convenience method for producing a NaN in IEEE-754 double precision (11 exponent bits, 53 significand bits).

These methods wrap `Z3_mk_fpa_nan`.


z3/src/ast.rs
```diff
@@ -1138,6 +1138,22 @@ impl<'ctx> Float<'ctx> {
...
+        unsafe { Self::wrap(ctx, Z3_mk_fpa_nan(ctx.z3_ctx, sort.z3_sort)) }
+    }
+
+    /// Convenience IEEE-754 single & double.
+    pub fn nan32(ctx: &'ctx Context) -> Float<'ctx> {
+        let s = Sort::float(ctx, 8, 24);
+        Self::nan(ctx, &s)
+    }
+    pub fn nan64(ctx: &'ctx Context) -> Float<'ctx> {
+        let s = Sort::float(ctx, 11, 53);
```
toolCHAINZ, line 1153: Why not use `Sort::double()`?

mehrad31415: updated.

toolCHAINZ:
Thanks for making these changes! Just one more tweak I'd like (this also applies to your other PRs):

I don't want to merge any new API without an accompanying set of unit tests, both to demonstrate the usage of the API and to make sure we're tracking it in our test suite.

Could you add some unit tests for these APIs (you'll find a giant list of tests for the rest of the project where you can add them; inline doc tests are even better)? Maybe something that confirms the expected behavior of NaN (not equal to other floats or itself)?

I also find having to immediately write code _using_ an API I write is good practice, as it forces me to confront how it will actually be used and whether improvements can be made.

toolCHAINZ:
Thanks for pushing in updates! Do note that some stuff has moved around in the repo (I split up the massive Ast file), so you'll need to pull master in and merge your changes into it

Issue #227, opened 2023-03-13
Use NonNull in Z3 safe bindings


Pat-Lafon:
As observed in #226, some of the implementations of the safe Z3 function calls check whether the pointers they get from the unsafe bindings are null. I have proposed that the safe bindings could rely on [NonNull](https://doc.rust-lang.org/std/ptr/struct.NonNull.html) to uniformly enforce this check.

To the best of my understanding, I'll use Model as an example:

The unsafe bindings for creating a Z3 Model return a raw pointer

```rust
#[doc(hidden)]
#[repr(C)]
#[derive(Debug, Copy, Clone)]
pub struct _Z3_model {
    _unused: [u8; 0],
}
/// Model for the constraints inserted into the logical context.
pub type Z3_model = *mut _Z3_model;

...

/// Retrieve the model for the last [`Z3_solver_check`]
///
/// The error handler is invoked if a model is not available because
/// the commands above were not invoked for the given solver, or if the result was `Z3_L_FALSE`.
pub fn Z3_solver_get_model(c: Z3_context, s: Z3_solver) -> Z3_model;
```

Model then gets exposed in the safe bindings as
```rust
/// Model for the constraints inserted into the logical context.
//
// Note for in-crate users: Never construct a `Model` directly; only use
// `Model::new()` which handles Z3 refcounting properly.
pub struct Model<'ctx> {
    ctx: &'ctx Context,
    z3_mdl: Z3_model,
}
```

```rust
unsafe fn wrap(ctx: &'ctx Context, z3_mdl: Z3_model) -> Model<'ctx> {
    Z3_model_inc_ref(ctx.z3_ctx, z3_mdl);
    Model { ctx, z3_mdl }
}

pub fn of_solver(slv: &Solver<'ctx>) -> Option<Model<'ctx>> {
    unsafe {
        let m = Z3_solver_get_model(slv.ctx.z3_ctx, slv.z3_slv);
        if m.is_null() {
            None
        } else {
            Some(Self::wrap(slv.ctx, m))
        }
    }
}
```

The observation made is that one has to remember to check and correctly handle the possibility of getting a null pointer from `Z3_solver_get_model`. Presumably there is an invariant that a Model struct only contains a valid pointer to a Z3 model. In some cases, the null check is more of a defensive programming mechanism since we might always expect that (a non-buggy usage of) Z3 will give a valid pointer, in other cases, null is a valid output of Z3 that should be handled.

[...]

PR #393, opened 2025-07-25
Warn When Attempting to Set 'ctrl_c' via Config

mehrad31415:
When setting `cfg.set_param_value("ctrl_c", "true");` We receive the following message:

```
WARNING: unknown parameter 'ctrl_c'
Legal parameters are:
  auto_config (bool) (default: true)
  ctrl_c (bool) (default: true)
  debug_ref_count (bool) (default: false)
  dot_proof_file (string) (default: proof.dot)
  dump_models (bool) (default: false)
  encoding (string) (default: unicode)
  model (bool) (default: true)
  model_validate (bool) (default: false)
  proof (bool) (default: false)
  rlimit (unsigned int) (default: 0)
  smtlib2_compliant (bool) (default: false)
  stats (bool) (default: false)
  timeout (unsigned int) (default: 4294967295)
  trace (bool) (default: false)
  trace_file_name (string) (default: z3.log)
  type_check (bool) (default: true)
  unsat_core (bool) (default: false)
  well_sorted_check (bool) (default: false)
```

It says `unknown parameter 'ctrl_c'` and then continues with `Legal parameters are: ...ctrl_c...`. The parameter should be set using `z3::set_global_param("ctrl_c", "true");`. For the time being this temporary fix at least prevents the ocntradictory warning message.

### DIFF
--- a/z3/src/config.rs
+++ b/z3/src/config.rs
@@ -39,6 +39,12 @@ impl Config {
     ///
     /// - [`Config::set_bool_param_value()`]
     pub fn set_param_value(&mut self, k: &str, v: &str) {
+        if k == "ctrl_c" {
+            eprintln!(
+                "WARNING: Parameter 'ctrl_c' is global-only; use `z3::set_global_param` instead."
+            );
+            return;
+        }
         let ks = CString::new(k).unwrap();
         let vs = CString::new(v).unwrap();
         self.kvs.push((ks, vs));


toolCHAINZ:
Out of curiosity, what version of `z3` are you linking against? I tried this (against `z3 4.14.1.0`) and did not get the same result:

```rust
use z3::{Config, Context};

fn main() {
    let mut cfg = Config::new();
    cfg.set_bool_param_value("ctrl_c", true);
    let _ = Context::new(&cfg);
}
```

Output:
```
WARNING: unknown parameter 'ctrl_c'
Legal parameters are:
  auto_config (bool) (default: true)
  debug_ref_count (bool) (default: false)
  dot_proof_file (string) (default: proof.dot)
  dump_models (bool) (default: false)
  encoding (string) (default: unicode)
  model (bool) (default: true)
  model_validate (bool) (default: false)
  proof (bool) (default: false)
  rlimit (unsigned int) (default: 0)
  smtlib2_compliant (bool) (default: false)
  stats (bool) (default: false)
  timeout (unsigned int) (default: 4294967295)
  trace (bool) (default: false)
  trace_file_name (string) (default: z3.log)
  type_check (bool) (default: true)
  unsat_core (bool) (default: false)
  well_sorted_check (bool) (default: false)
```

Since this seems to vary between versions, I am hesitant to add this in as that might be more misleading.

mehrad31415:
I am using Z3 version `4.15.2.0` and get that result.

toolCHAINZ:
Since that option does not seem to exist in older versions of Z3, I think I would want this to first check if the version is new enough to suggest this. I could see it being frustraiting if the bindings pointed you to a non-existent flag if you're using an older z3 (which I think many are)

mehrad31415:
True, but where can I find out from which version of Z3 the option was removed? The simple approach is to just hardcode the version where the inconsistency appears (`4.15.2.0`), and assume it’s removed in that version and all later ones.

toolCHAINZ:
I think you would look at the Z3 repo and find the commit where the param was added (it hasn't been removed unless I'm misunderstanding something). 

However, I'm not totally sold on the strategy of printing an error message, ideally failing to set a parameter should be reflected in a `Result` type, so I think this is better put off to #227 when we add `Result` to fallible stuff.

PR #398, opened 2025-07-26
feat: add BV::from_bits 

Evian-Zhang:
If we have a biginteger type, it is usually implemented as `Vec<u64>` internally. To create a `BV` from such instance, converting to string and then letting Z3 parse it is a bad style for me. The `from_bits` method can avoid such printing and parsing.


toolCHAINZ:
Could you add a unit test for this API demonstrating its use? Maybe show equivalence to a BV constructed through another method, just something that exercises it.

Evian-Zhang:
Added an example in docs for doctests.

PR #401, opened 2025-08-04
changed!: refcount z3 context

toolCHAINZ:
This PR makes `Context` internally refcounted on the Rust side with an `Rc` around a `pub(crate)` wrapper implementing `Drop`. This has a couple effects:

* All other z3 types (e.g. all `Ast`s, `Solver`s, etc) _own_ their (refcounted) `Context` handle, removing almost all instances of the `'ctx` lifetime (with the exception of those associated with the `ContextHandle` type).
* The `Context` is refcounted on the rust side any time an `Ast`/`Solver`/etc is cloned and decremented any time it is dropped.

Simple consuming code will mostly be unchanged due the compiler eliding lifetimes in the old version. Removing the occasional lifetime argument was the only change required for the existing test suite.

This is a *BREAKING* change: consuming code with types wrapping `z3.rs` types must be migrated to remove the `'ctx` lifetime.

So:

```rust
struct MyStruct<'ctx>{
    a: BV<'ctx>,
    b: BV<'ctx>,
}
```

becomes:

```rust
struct MyStruct{
    a: BV,
    b: BV,
}

### DIFF
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -16,69 +16,69 @@ use crate::{Context, FuncDecl, IsNotApp, Pattern, Sort, SortDiffers, Symbol};
 use num::{bigint::BigInt, rational::BigRational};
 
 /// [`Ast`] node representing a boolean value.
-pub struct Bool<'ctx> {
-    pub(crate) ctx: &'ctx Context,
+pub struct Bool {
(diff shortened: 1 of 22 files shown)


wtdcode:
This is cool and solves our issue =).

PR #102, opened 2020-10-30
[WIP] Send for SendableContext (unsafe, but clearly marked as unsafe)

Trolldemorted:

Issue #100, opened 2020-10-28
Aren't Config, Context, Solver and friends Send-able?


Trolldemorted:
As @hengchu mentioned [here](https://github.com/prove-rs/z3.rs/pull/48#issuecomment-515810062), it is not obvious why these types should not implement `Send`. Just because those types aren't safe for concurrent usage, they are not necessarily unsafe to be used by different threads sequentially.

Them being !Send makes it impossible to use them in worker queue scenarios, which is a serious restriction of the usability.

Do you have an idea whom one could ask or which documentation one could read about that?

Trolldemorted:
I have raised an upstream issue for that, since the only discussion on the topic I could find was >5 years old.

https://github.com/Z3Prover/z3/issues/4762

Trolldemorted:
Apparently context + all its expressions can be moved to a different thread. Do you have an idea how we could enforce that *everything* is moved to a different thread in one go, and not just a single AST/Context/Foo object?

fitzgen:
Thanks for checking with upstream.

I suppose we could dynamically keep track of the thread id for the context, and then every operation would get the context and assert that it is currently on the context's thread... That seems really heavy though. I'm not convinced that would actually be an overall improvement to the library vs the current state.

Trolldemorted:
I have opened a PR with a simpler proposal:

- You can call `prepare_move` (unsafely) with a context, a set of asts, and a set of solvers
- In there, everything is marked as moved
- It returns a `SendableContext` which contains the raw pointers and is declared Send
- You may rebuild your context with ` pub unsafe fn from_raw(z3_ctx: *mut z3_sys::_Z3_context) -> Context`
- You may rebuild your solver with `pub unsafe fn from_raw(ctx: &'ctx Context, z3_slv: *mut z3_sys::_Z3_solver) -> Solver<'ctx>`
- You may rebuild your asts with `fn new(ctx: &'ctx Context, ast: Z3_ast) -> Self;` (Shouldn't this be unsafe? In this situation it clearly is unsafe)
- All drop implementations honor the moved marker, and don't double free

If you like it I'd propose we also add `asserts!(!self.moved)` all over the place, and track whether the elements in `SendableContext` are being reconstructed more than once - that should reduce the dangerousness of the API from "if you don't honor the unsafe requirements of `prepare_move` you will corrupt your memory" to "if you don't honor the unsafe requirements of `prepare_move` you will get a panic".

toolCHAINZ:
[...]

PR #404, opened 2025-08-06
feat: enable moving and referencing z3 types between threads

toolCHAINZ:
This is an attempt at making z3.rs usable with threading (see also #100, #102). The basic strategy is to make a wrapper type that owns both a context, and some piece of data (or collection of pieces of data) relying on that context (e.g. an Ast, a Solver, a Model). This handle is the only user of its context and so it and all its Asts are able to be moved to another thread, and then unwrapped using that thread's context.

Here's an example from a unit test:

```rust
#[cfg(test)]
mod tests {
    use crate::Context;
    use crate::ast::Bool;
    use crate::translate::{PrepareSendable, RecoverSendable};

    #[test]
    fn test_send() {
        let ctx = Context::default();
        let b = Bool::from_bool(&ctx, true);
        // create sendable handle
        let sendable = b.prepare_sendable(); // sendable: SendableHandle<Bool>
        std::thread::spawn(move || {
            let thread_ctx = Context::default();
            // the sendable handle is moved across threads
            // the `recover` function allows for extracting to Bool back out
            let moved = sendable.recover(&thread_ctx); // moved: Bool
            assert_eq!(moved.as_bool(), Some(true));
        })
        .join()
        .expect("uh oh");
    }
}
```

So far, I've only implemented `Ast`s, but it should be easily extendable to all the other types.

This method makes it possible to make a `SendableHandle` for both an individual piece of data and for a collection. So you can also do `.prepare_sendable()` on `&[T]` and `Vec<T>` where `T` is an `Ast` thanks to a default trait impl. Doing it like this makes _1_ context for the whole thing, not one-per-item as I had originally feared would be necessary.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -37,7 +37,6 @@ jobs:
       run: cargo test --workspace
 
   build_on_wasm:
-    if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -37,6 +37,7 @@ jobs:
       run: cargo test --workspace
 
   build_on_wasm:
+    if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
--- a/z3/Cargo.toml
+++ b/z3/Cargo.toml
@@ -16,7 +16,7 @@ edition = "2024"
 
 
 [features]
-default = []
+default = ["synchronization"]
 bundled = ["z3-sys/bundled"]
 vcpkg = ["z3-sys/vcpkg"]
 gh-release = ["z3-sys/gh-release"]
--- a/z3/Cargo.toml
+++ b/z3/Cargo.toml
@@ -16,6 +16,7 @@ edition = "2024"
 
 
 [features]
+default = []
 bundled = ["z3-sys/bundled"]
 vcpkg = ["z3-sys/vcpkg"]
 gh-release = ["z3-sys/gh-release"]
--- a/z3/Cargo.toml
+++ b/z3/Cargo.toml
@@ -16,11 +16,9 @@ edition = "2024"
 
 
 [features]
-default = ["synchronization"]
 bundled = ["z3-sys/bundled"]
 vcpkg = ["z3-sys/vcpkg"]
 gh-release = ["z3-sys/gh-release"]
-synchronization = []
 
 # This is a legacy feature here for short term compatibility.
 static-link-z3 = ["z3-sys/bundled", "z3-sys/deprecated-static-link-z3"]
@@ -34,6 +32,7 @@ num = "0.4"
 [dev-dependencies]
 env_logger = "0.11"
 semver = "1"
+rayon = "1.10.0"
 
 [dependencies.z3-sys]
 path = "../z3-sys"
--- a/z3/Cargo.toml
+++ b/z3/Cargo.toml
@@ -20,6 +20,7 @@ default = []
 bundled = ["z3-sys/bundled"]
 vcpkg = ["z3-sys/vcpkg"]
 gh-release = ["z3-sys/gh-release"]
+synchronization = []
 
 # This is a legacy feature here for short term compatibility.
 static-link-z3 = ["z3-sys/bundled", "z3-sys/deprecated-static-link-z3"]
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -415,6 +415,10 @@ pub trait Ast: fmt::Debug {
     }
 }
 
+pub trait Translate: Ast {
+    fn translate(&self, dest: &Context) -> Self;
+}
+
 macro_rules! impl_ast {
     ($ast:ident) => {
         impl Ast for $ast {
@@ -445,8 +449,8 @@ macro_rules! impl_ast {
             }
         }
 
-        impl $ast {
-            pub fn translate(&self, dest: &Context) -> $ast {
+        impl Translate for $ast {
+            fn translate(&self, dest: &Context) -> $ast {
                 unsafe {
                     $ast::wrap(dest, {
                         Z3_translate(self.get_ctx().z3_ctx.0, self.get_z3_ast(), dest.z3_ctx.0)
--- a/z3/src/ast.rs
+++ b/z3/src/ast.rs
@@ -13,8 +13,8 @@ use z3_sys::*;
 
 use crate::{Context, FuncDecl, IsNotApp, Pattern, Sort, SortDiffers, Symbol};
 
-use num::{bigint::BigInt, rational::BigRational};
 use crate::translate::Translate;
+use num::{bigint::BigInt, rational::BigRational};
 
 /// [`Ast`] node representing a boolean value.
 pub struct Bool {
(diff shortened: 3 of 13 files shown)


Pat-Lafon:
Out of curiosity, can you create the objects once, create sendable objects, and then create N threads, and in each thread receive these objects? It's kinda cool if you can use this to run multiple variations of the same query with different solver parameters for each thread and such(I think this is the rayon model).

toolCHAINZ:
Hmm, I hadn't considered that. I think reusing a Sendable object would be possible with a customized Clone impl. The Clone would have to create a new context for each new clone (and translate its payload)to remain sound I think.

toolCHAINZ:
I simplified the implementation further. There's now two traits in play:

`Translate` is a standalone trait for things that can be translated from one context to another. I have reused the existing `translate` impls and moved them into `Translate` implementations. This has been done for all types with `translate`: `Goal`, `Solver`, `Model`, and a blanket impl for all types that implement `Ast`. I have also implemented it for `Vec<T> where T: Translate`. 

`Translate` is marked as `unsafe` because it relies on implementations actually translating their contents to the provided context.

`PrepareSendable` is a safe trait for making a `SendableHandle`. `SendableHandle` can only be constructed outside this crate with `SendableHandle::new`, which only accepts types that are `Translate`, and has a fixed implementation that ensures the types are translated into the handle's context (the safety of this assumes the soundness of the type's `Translate` impl).

I've defined a blanket impl of `PrepareSendable` for all types that are `Translate`. I've also done a specialized implementation for `&[T]` (since `&[T]` cannot implement `Translate`) in which I manually construct the `SendableHandle` (still ensuring the above requirements are met).

Users can use this system with their own `z3`-derived types by implementing `Translate`, and they can just ignore `PrepareSendable`.

Here's an example of a round-trip of some Z3 stuff to a thread:

```rust 
    #[test]
    fn test_round_trip() {
        let ctx = Context::default();
        let bool = Bool::new_const(&ctx, "hello");
        let sendable = bool.prepare_sendable();
        let model = std::thread::spawn(move || {
            let thread_ctx = Context::default();
            let moved = sendable.recover(&thread_ctx);
            let solver = Solver::new(&thread_ctx);
            solver.assert(&moved._eq(&Bool::from_bool(&thread_ctx, true)));
            solver.check();
            let model = solver.get_model().unwrap();
            model.prepare_sendable()
        })
        .join()
        .expect("uh oh");
        let model = model.recover(&ctx);
        assert_eq!(model.eval(&bool, true), Some(Bool::from_bool(&ctx, true)));
    }
```

(Oddly, I did this with `Bool::fresh_const` earlier and this test failed in CI (only on linux) saying that the model of `bool` was false. This might be a z3 bug, or I'm misunderstanding something? Changing it to `new_const` seemed to stop the error; very strange)

Another note, I suspect this stuff won't work with `wasm`. We may want to gate this behind a feature flag as I'm not sure if it will build. The tests certainly won't work.

I added `Clone` to `SendableHandle`, so the rayon use-case _should_ work. I have yet to try it though.

toolCHAINZ:
I just gave a quick try with `rayon` and I don't think it will work in the rayon paralleism model. `rayon` requires that closure being run in parallel itself to be `Send` and `Sync`. `SendableHandle` is not `Sync` (to ensure we don't reference it across threads). ~And even if it were, we have to unwrap it to use the inner data, and that inner data will be `!Send` and `!Sync`.~

I think it _might_ work in an async setup like `async-executor` which allows for non-work-stealing workloads, but I have not tried this.

toolCHAINZ:
Actually, I take that back. For a laugh, I tried implementing `Sync` on `SendableHandle` and it compiled. So I guess unwrapping it is not a problem. I could get the following code to compile and it runs without issue:

```rust
    #[test]
    fn test_rayon() {
        use rayon::prelude::*;
        let ctx = Context::default();
        let int = Int::new_const(&ctx, "hello");
        let sendable = int.prepare_sendable();
        (0..100000).into_par_iter().for_each(|i| {
            let ctx = Context::default();
            let moved = sendable.recover(&ctx);
            let solver = Solver::new(&ctx);
            solver.assert(&moved._eq(&Int::from_i64(&ctx, i * 2)));
            assert_eq!(solver.check(), crate::SatResult::Sat);
            let model = solver.get_model().unwrap();
            assert_eq!(model.eval(&moved, true), Some(Int::from_i64(&ctx, i * 2)));
        })
    }
``` 

So that seems like it works? I'm a little nervous about implementing `Sync` on `SendableHandle`. I feel like this is opening up UB in some way but I can't put my finger on it. My mental model for the safety here was that we were sure that the contexts inside `SendableHandle` could only ever be accessed inside the same thread as the `SendableHandle`.

I'm kind of puzzled that example works. It's even not `move`-ing handles into the closure.

My (maybe unfounded) suspicion: it is safe to access a z3 `Context` across threads without synchronization as long as no new declarations are made inside that `Context`. Since all the creation inside the handle's `Context` happens when it the handle is initially created, and we have made it impossible to ever allocate anything else inside it, is is effectively "frozen" and can be referenced across threads safely. The refcount on the Context also ensures that it lives as long as anything references that `SendableHandle`.

I should also note, 20-30% of the run time in that snippet is spent in the `Context::default()` in each individual iteration. Practical usage would probably use a worker pool model or a thread-local static context to avoid this repeated allocation.

toolCHAINZ:
I played around with this and confirmed that, if the context has additional objects declared, it leads to instant segfaults (I tested this by declaring a `Bool` in the handle's context any time something was `recover`'d.

toolCHAINZ:
Ok, following some guidance in https://github.com/Z3Prover/z3/discussions/7762 I think I'm going to make the handle `Send` and `Sync`, but put a `Mutex` in it. Since we are quarantining the `Context` and `Ast` inside the struct, we do _not_ have to put any mutex on any other types, and I think given the API here there's no way for this to starve or deadlock.

Pat-Lafon:
Wow, very cool work. I don't think I can give you much advice on the unsafe code side of things. Just to clarify, since you feature gated this, do the synchronization tests not run in ci?

toolCHAINZ:
I feel comfortable with the safety of this implementation now with the Mutex added. I've removed the rayon test to avoid adding the dependency but verified that this Mutex'd implementation does work just like the example above.

Yeah I'm currently trying to think through the feature thing. Should this be a default feature?

If we make it default we will probably need to change wasm CI to not build it.

If it's not default I guess we change the default build job to specifically activate it to make sure it's tested.

I lean towards making this default?

toolCHAINZ:
Ok, that's surprising; I had enabled wasm CI to test the feature stuff, but as far as I can tell `wasm` is building with no issues. Maybe the feature flag is unneeded and it _can_ compile it?

toolCHAINZ:
Playing around with `wasm`, it looks like code using threading primitives _does_ compile but then hits runtime issues. I'm now on the fence about whether to feature gate this at all.

If it's not behind a feature, then it's basically just code that has no practical usage but is exposed to wasm programs.

Pat-Lafon:
> I feel comfortable with the safety of this implementation now with the Mutex added. I've removed the rayon test to avoid adding the dependency but verified that this Mutex'd implementation does work just like the example above.

Note you can include it as a dev-dependency but fair enough.

> Yeah I'm currently trying to think through the feature thing. Should this be a default feature?
> 
> If we make it default we will probably need to change wasm CI to not build it.

Is there a way to gate this on non-wasm setups? Looking at the docs, something like `#[cfg(not(target_family = "wasm"))]`?

toolCHAINZ:
Wow, I completely forgot that `dev-dependencies` were a thing. 😆 

Thank you for the reminder, I've added it as a dev dep and re-added the test.

I also looked around a little more and found that the threading story on `wasm` is more nuanced that I thought. While there's no `std::thread`, it's still possible to define threading primitives (e.g. https://github.com/RReverser/wasm-bindgen-rayon), so these `Send`/`Sync` types are still applicable.

Therefore, I've removed the `synchronized` feature and merely gated the _tests_ that use `std::thread` from being compiled on `wasm`.

PR #417, opened 2025-08-19
feat: Use a default z3 context

### LINKED ISSUE #261: Use an "implicit" `Context`?
It is tempting to move to using a `thread_local!(static CONTEXT: Option<z3::Context> = None)` and some helper functions to have a scoped Context ... this would let us remove a bunch (all?) of the `Context` parameters to the various functions in the `z3` crate.

This would help make the API look more like the Python or JS APIs, while still letting people have multiple contexts if they need them.

toolCHAINZ:
Closes #261 

This seemed like the next natural step after some recent refactors.

Z3.rs has several functions that construct types which require a Context. These functions all have a `&Context` argument. The strategy pursued in this MR, to reduce the footprint of this change, is to rewrite all these functions from `<name>(ctx: &Context, ...)` to `name_in_context(ctx: &Context, ...)` and to introduce a new wrapper, with the original name but without the context argument: `<name>(...)`.

These outer wrapper functions can then use a newly defined `Context::thread_local()` function to get a handle to a "default" thread-local `Context` and provide that to the inner function. This thread-local default context can also be replaced with one generated from a provided `&Config` (see the `unknown` Solver and Optimize tests).

This PR pursues this strategy by way of an attribute macro `#[z3(Context::thread_local)]`, where the `Context::thread_local` is telling the macro what function to call to get a handle to the context. The attribute macro provides the above rewriting automatically for both individual `fns` and for whole `impl` blocks, reducing the noise in the code. Any function taking a `&Context` argument in an `impl` block decorated with this attribute macro will be automatically rewritten as in the above description.

The actual meaningful changes in this MR are definition of the attribute macro and its application to all relevant attribute blocks. The rest of the changes are just tests and doc tests being fixed to use the new default context settings (other than ones specifically testing multi-context stuff).

There are a few functions that I haven't ported here (e.g. lambdas, quantifiers), as they're outside impl blocks and the macro is designed with usage inside an impl block in mind, and I'd rather keep it as simple as possible.


toolCHAINZ:
Following some discussion (https://github.com/prove-rs/z3.rs/discussions/400#discussioncomment-14174134) I've reworked this from the initial description.

This PR still defines a `thread_local` Context, and the public API is still changed to remove all `ctx` arguments (with the implementation using the default `thread_local` context. 

Now, however, there are not macro-generated `in_ctx` variants of these APIs for the use-case where a user wants to customize their context. Instead, we introduce a top-level crate functions: `z3::with_z3_context` (and a "config" helper variant). This function accepts a context and a closure. It replaces the thread-local context with this context and then runs the closure. It replaces the original thread-local context after the closure has completed and returns the data it returned.

The closure and the return type are both required to be `Send` and `Sync`, which provides the handy guarantee that Z3 types cannot escape or enter the closure, except through the `Synchronized` wrapper. I believe this statically prevents all "context mismatches" in safe code.

Here's an example from a test:

```rust
#[test]
fn test_solver_unknown() {
    let _ = env_logger::try_init();
    let mut cfg = Config::new();
    // Use a very short timeout to quickly return "unknown"
    cfg.set_timeout_msec(1);
    with_z3_config(&cfg, || {
        // An open problem: find a model for x^3 + y^3 + z^3 == 42
        // See: https://en.wikipedia.org/wiki/Sums_of_three_cubes
        let x = ast::Int::new_const("x");
        let y = ast::Int::new_const("y");
        let z = ast::Int::new_const("z");
        let x_cube = ast::Int::mul(&[&x, &x, &x]);
        let y_cube = ast::Int::mul(&[&y, &y, &y]);
        let z_cube = ast::Int::mul(&[&z, &z, &z]);
        let sum_of_cubes = x_cube + y_cube + z_cube;
        let sum_of_cubes_is_42 = sum_of_cubes._eq(42);

        let solver = Solver::new();
        solver.assert(&sum_of_cubes_is_42);

        assert_eq!(solver.check(), SatResult::Unknown);
        assert!(solver.get_reason_unknown().is_some());
    });
}
```

This test needs to run in a context with a timeout. It does so by running the tests insde the closure.

Here's another one:

```rust
#[test]
fn test_solver_translate() {
    let a = ast::Int::new_const("a");
    let s = a.synchronized();
    let slv = with_z3_config(&Config::new(), || {
        let a = s.recover();
        let slv = Solver::new();
        slv.assert(a._eq(2));
        assert_eq!(slv.check(), SatResult::Sat);
        slv.synchronized()
    })
    .recover();
    slv.assert(a._eq(3));
    assert_eq!(slv.check(), SatResult::Unsat);
}
```

In this test, we move a `Synchronized<Int>` into the closure and we move a `Synchronized<Solver>` out. This works just like it does for threads.

PR #420, opened 2025-08-23
doc: A first pass at documenting how to define recursive datatypes

### LINKED ISSUE #419: Creating recursive datatypes
I am wondering how I can create recursive datatypes because the datatype is supposed to refer to itself, so I do not see how the index based approach would fit in with it.

https://microsoft.github.io/z3guide/docs/theories/Datatypes/#recursive-datatypes

Pat-Lafon:
Fixes #419 

I did a rough pass on how I think one is supposed to define recursive sorts(both a single sort and mutually recursive sorts). I've implemented it as a doc test so there is some checking, but I didn't attempt to invoke the solver here(there is supposedly a proper test for that).

I don't love the way the docs are shown in docs.rs since this module only shows the create_datatypes function and the DatatypeBuilder is on a separate page. (The new documentation would go here https://docs.rs/z3/latest/z3/datatype_builder/index.html where as the builder is on the top level https://docs.rs/z3/latest/z3/struct.DatatypeBuilder.html ).

Open to any suggestions(or one can make the changes themself, I won't be offended).

### DIFF
--- a/z3/src/datatype_builder.rs
+++ b/z3/src/datatype_builder.rs
@@ -12,8 +12,8 @@
 //! ```rust
 //! use z3::{Sort, DatatypeAccessor, DatatypeBuilder, Symbol};
 //! let dt = DatatypeBuilder::new("my_datatype")
-//!     .variant("case1", vec![("field1", DatatypeAccessor::Sort(Sort::int()))])
-//!     .variant("case2", vec![("field2", DatatypeAccessor::Datatype(Symbol::String("my_datatype".to_string())))])
+//!     .variant("case1", vec![("field1", DatatypeAccessor::sort(Sort::int()))])
+//!     .variant("case2", vec![("field2", DatatypeAccessor::datatype("my_datatype"))])
 //!     .finish();
 //! ```
 //!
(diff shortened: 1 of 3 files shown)


toolCHAINZ:
This looks great to me! Maybe we should do a breaking change at some point and re-organize how things are exported. I think there's a few other mods/types that exhibit this kind of "docs.rs-hostile" layout.

I took the liberty of pushing in a couple additions and tweaks. I added new methods to produce DataTypeAccessors using `Into<Sort>` and `Into<Symbol>`. This cuts down on a lot of parentheses bloat in the new doc example. I made the methods for producing the `DataTypeAccessor` attached to the `DataTypeAccessor` type but IDK if they would be better attached to the `DataTypeBuilder` or `DataType` types instead.

If you could look over those changes and see if they seem sane (and change whatever you need to) that'd be great as I'm not familiar with the API.

I propose we merge it now as-is and think about reorganizing things/docs later.

toolCHAINZ:
Maybe `DataType::by_sort` or `DataType::by_symbol` would be a clearer way to make accessors?

Pat-Lafon:
Your changes look good to me! 

I guess the accessor name is more closely tide to the underlying implementation. A higher-level name would be better(only worried that `DataType::by_sort` might make is sound like this is creating the datatype? maybe just uncertainty on my end)

toolCHAINZ:
Sounds good! I'll merge as is now and we can think about if there's a better name or structure for this stuff later.

PR #425, opened 2025-08-26
feat(z3-sys): do not depend on std

lucascool12:
This PR removes z3-sys' dependency on std.
This gives users of z3-sys the choice to pull in std or not, which is currently not possible.

### DIFF
--- a/z3-sys/src/lib.rs
+++ b/z3-sys/src/lib.rs
@@ -30,6 +30,7 @@
 #![allow(non_camel_case_types)]
 #![allow(clippy::unreadable_literal)]
 #![warn(clippy::doc_markdown)]
+#![no_std]
 
 mod generated;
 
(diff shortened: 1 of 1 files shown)


toolCHAINZ:
Just to establish the linkage, wanted to note that this (in general) was brought up here: https://github.com/prove-rs/z3.rs/discussions/372#discussioncomment-13784382, though that was in the context of making both `z3` and `z3-sys` `no-std`. Following #401 the best we could do for the `z3` crate would be `core + alloc` anyway.

Out of curiousity, do you have a specific use-case in mind for this? Is it just for size optimization? I have trouble thinking of a target that could handle Z3 (quite allocation and CPU hungry) but not a `std` rust executable so I'm curious if there's a specific application I'm not thinking of.

All that said, since `z3-sys` is such a low-level wrapper and it has so little going on, I don't see a compelling reason _not_ to do this. We would like to do some refactoring to it (#227) to be a little more idiomatic (e.g. return `Option<NonNull<T>>` instead of `*mut T`) but those changes would still work in `no-std`.

Thanks for submitting this!

toolCHAINZ:
I'm still curious to hear the answer, but will go ahead and merge this. Thanks again!

lucascool12:
> I'm still curious to hear the answer, but will go ahead and merge this. Thanks again!

The specific use case I have in mind is for reducing wasm build sizes.
The impact will probably not be incredible because z3 is quite chunky, but every bit counts.

PR #436, opened 2025-09-06
changed!: APIs accepting `Bool` no longer accept `bool`

toolCHAINZ:
Follow up to #305 and #410 to fix some ergonomic issues pointed out by @dragazo.

In #410 (and a follow-up PR later), I changed `Solver::assert()` to take `Into<Bool>` instead of `&Bool`. The desire here was to remove the need to throw `&` on `Bool` arguments even if you were fine consuming them (e.g. allow `solver.assert(a.eq(b))` instead of needing `solver.assert(&a.eq(b))`.

Unfortunately, due to the existance of `impl Into<Bool> for bool`, this meant that you can also pass raw `bool`s in. This is something that is not commonly needed and also opens the door for something like this: `solver.assert(a == b)`, which is using the `Eq` impl, which itself uses a syntactic "ast equality" check to immediately return a `bool`. This _looks_ like adding an assertion to the solver, but is actually just adding `true` or `false`, so never something you actually want.

However, I do want to keep the ability to consume `Bool`s passed into assert, so instead of just reverting to `&Bool`, I'm accepting `Borrow<Bool>`.

This passes the entire test suite (which has both `assert(&b)` and `assert(b)` forms) with no changes needed other than fixing one test that used this functionality to explicitly do `Solver::assert(false)` to test UNSAT results. 

With this change the compiler now prevents `solver.assert(a == b)` with the message:
> Trait `Borrow<Bool>` is not implemented for `bool` 

This message at least explicitly names both types which can hopefully clue in someone misusing this function.

Since this now prevents giving `bool` to `assert` (even though doing so was likely only ever done in error), this counts as a breaking change.

### DIFF
--- a/z3/src/ast/bool.rs
+++ b/z3/src/ast/bool.rs
@@ -145,15 +145,3 @@ impl Bool {
         }
     }
 }
-
-impl From<bool> for Bool {
-    fn from(value: bool) -> Self {
-        Bool::from_bool(value)
-    }
-}
-
-impl From<&bool> for Bool {
-    fn from(value: &bool) -> Self {
-        Bool::from_bool(*value)
-    }
-}
--- a/z3/src/ast/float.rs
+++ b/z3/src/ast/float.rs
@@ -256,7 +256,7 @@ impl_into_ast!(f64, from_f64);
 #[cfg(test)]
 mod tests {
     use crate::Solver;
-    use crate::ast::{Ast, Float};
+    use crate::ast::{Ast, Bool, Float};
 
     #[test]
     fn test_nonstandard_float() {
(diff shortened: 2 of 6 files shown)


toolCHAINZ:
Hmmm, it occurs to me that this will stop `assert(a == b)` but not `assert(a.eq(b == c))`. I think you're righ @dragazo, I need to take a longer look at the `Bool` stuff; likely `Bool` just needs to be special and needs to have slightly different implementations than the other `Asts`, specifically to only accept only actual `Bool`s when building `Bool` asts.

toolCHAINZ:
Ok, I removed `impl From<bool> for Bool`. This makes it so that `==` cannot be used at all in anything that accepts a `Bool`.

There's a slight cost, as seen in the unit test changes here, of building expressions with concrete `bools` being more verbose, but I don't think these were a big deal? Like any expression with a concrete `true` or `false` in it can be simplifed anyway, so we can just think of this as a gentle nudge to the user not to make redundant formulae.

PR #437, opened 2025-09-06
doc: add simple top-level example

toolCHAINZ:


iuaphe:
Hey! Just to let you know, there's a small typo in the example. The code asserts that Alice's has 3 *times* as many apples as Charlie, but the English description of the constraint is:

> - The number of apples Alice receives must be exactly ***3 more than*** the number Charlie receives.

(my emphasis)

Also, for the "at least one" checks, I think you should use `gt(0)`, not `ge(0)`.

toolCHAINZ:
Thank you for the proof read! I will fix this!

toolCHAINZ:
I fixed the doc typos in https://github.com/prove-rs/z3.rs/pull/448

PR #440, opened 2025-09-08
fix: Solver::clone now preserves tactics and params

toolCHAINZ:


Pat-Lafon:
Very nice, I've gotten bitten by the weird nature of Z3 parameters before. I do wish there was a good way to print these out(even just for logging purposes for more reproducible queries). Like when I use to_string on a Solver, i don't get any additional setting that might have been set like rlimit

toolCHAINZ:
Yeah, totally agreed, I was shocked there was no way to inspect or log any of that.

PR #452, opened 2025-09-26
Use FnOnce for closures that are run once

gmorenz:
We only use these closures once, so they can be `FnOnce` closures that consume their captures.

I specifically want this because I build a string from some non `Sync` data and then pass that to `solver.from_string`, so without this I need to clone that string for no reason.


toolCHAINZ:
This is a great catch, thanks!

toolCHAINZ:
Setting this to release as `0.19.1`. Thanks again!

PR #455, opened 2025-10-20
feat: Add `Datatype::update_field`, `FuncDecl::domain`, `FuncDecl::range`

willcrichton:
I was playing around with datatypes but noticed that `update-field` isn't exposed, so I added a datatype method to implement it. I also added `domain` and `range` functions to `FuncDecl` which are used to check preconditions on the update method, but should also be more generally useful. I've included tests for all methods.

### DIFF
--- a/z3/src/optimize.rs
+++ b/z3/src/optimize.rs
@@ -283,6 +283,12 @@ impl Optimize {
     }
 }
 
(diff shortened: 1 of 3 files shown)


toolCHAINZ:
Looks good, thanks!

PR #464, opened 2025-11-14
feat: Implement support for bundling z3 without use of github checkout

ThomasTNO:
This PR implements support for bundling Z3 even if you do not specify `z3` as git dependency.

### DIFF
--- a/z3-sys/build.rs
+++ b/z3-sys/build.rs
@@ -322,11 +322,12 @@ fn generate_binding(header: &str, search_paths: &[PathBuf]) {
     }
 }
 
-/// Build z3 with bundled source codes.
 #[cfg(feature = "bundled")]
 use gh_release::{download_unzip, get_github_client};
+#[cfg(feature = "bundled")]
 use reqwest::blocking::Client;
 
+#[cfg(feature = "bundled")]
 fn get_z3_submodule_url(client: &Client, z3_sys_version: &str) -> String {
     let submodule_extract_url = format!(
         "https://api.github.com/repos/prove-rs/z3.rs/contents/z3-sys/z3?ref=z3-sys-v{z3_sys_version}"
@@ -349,6 +350,7 @@ fn get_z3_submodule_url(client: &Client, z3_sys_version: &str) -> String {
         + ".zip"
 }
 
+#[cfg(feature = "bundled")]
 fn build_bundled_z3() {
     let z3_sys_version = env!("CARGO_PKG_VERSION");
     let z3_dir = PathBuf::from(env::var("OUT_DIR").unwrap()).join(format!("z3"));


ThomasTNO:
@toolCHAINZ, let me know if any modifications are needed.

toolCHAINZ:
@Pat-Lafon @lmondada any thoughts on this?

I personally never use this feature. The change seems reasonable to me though. As you've both had opinions in the past and contributed related features I'm curious about your opinion.

lmondada:
Hi, thanks for looping me in!

Do I understand correctly that this would download the library from GitHub? (i.e. `bundled` would act just like `gh-release`)?

I personally find the gh-release solution super useful and use it always, but am not sure about doing this without the user's explicit concent because:

- it comes with a lot of additional build dependencies 
- in certain use cases (e.g. CI), GitHub rate limiting is a very real & annoying fact of life
- it's a bit "hacky", in the sense that I'm not sure this is the "right way" of obtaining copies of Z3

I wish there was a better way of bundling Z3 that didn't involve scraping GitHub...

toolCHAINZ:
Yeah, I believe the intent here is to have bundled download the z3 source repo (needed to do a static build) with similar logic to gh-release instead of using the git submodule checkout (which does not exist when used as a crates.io dep).

I do wonder if the build options are misnamed at this point. I could see them decomposed maybe:
- default = dynamic build + existing dynamic library 
- gh-release = dynamic build + fetched z3 dynamic library 
- bundled = static build + fetched and built z3 static library 
All the env variables are then basically just setting various paths to customize one of those scenarios.

Maybe in the future we should have a static / dynamic feature and a "how to get z3" feature or something

ThomasTNO:
Alternatively we might be able to package the entire Z3 with the crate, but in that case anyone not requiring a static build is downloading a copy of Z3 within the crate.

I see there exists a crate to include a directory through a macro: https://docs.rs/include_dir/latest/include_dir/

The docs suggest quite an impact on performance though..

ThomasTNO:
> Hi, thanks for looping me in!
> 
> Do I understand correctly that this would download the library from GitHub? (i.e. `bundled` would act just like `gh-release`)?
> 
> I personally find the gh-release solution super useful and use it always, but am not sure about doing this without the user's explicit concent because:
> 
> * it comes with a lot of additional build dependencies
> * in certain use cases (e.g. CI), GitHub rate limiting is a very real & annoying fact of life
> * it's a bit "hacky", in the sense that I'm not sure this is the "right way" of obtaining copies of Z3
> 
> I wish there was a better way of bundling Z3 that didn't involve scraping GitHub...

It will only download Z3 if it is not available yet and your going for a static build. You can circumvent the github scraping and download by specifying the location of a pre-downloaded copy of Z3. Something that was as a matter of fact already required. I would argue this is not a breaking change.

toolCHAINZ:
Yeah I agree this shouldn't be a breaking change. The bundled feature doesn't even work with crates.io so something that makes it work seems an obvious improvement to me.

I'm trying to think if anything unexpected could happen with this setup. If someone were to check out the repo but not init submodules would this behave reasonably?

Before I merge this, when I'm back at a computer, I'd like to mirror this onto a branch directly on the z3.rs repo to force the bundled CI jobs to run.

ThomasTNO:
> If someone were to check out the repo but not init submodules would this behave reasonably?

Good question. It should download the version that's available on GitHub by default, regardless of the existence (init) of the submodule. So the behaviour changes a bit in that case since currently it would attempt to use the submodule (which is not there).

toolCHAINZ:
I've made #466 to force the CI to run for these changes. I've also altered the CI so that it doesn't check out the submodule to try to simulate the "no submodule" case and the crates.io case. If that CI all passes, I think I'm fine merging this.

I don't love the implicit dependency on the z3.rs git repo, the submodule location, or the tag format, but this is a kind of weird building situation and I can't think of a cleaner way to do it. So I'm happy to try this out for now and see if it works well.

Since the tags are automatically generated by `release-plz` it's at least not totally brittle.

toolCHAINZ:
Ok, the CI on the other PR looks good. I'm going to merge this; thanks again for the contribution!

PR #471, opened 2025-11-20
refactor: remove unused imports in z3::ast

lixitrixi:
Clippy warns about unused imports `IntoAst` and `binop` in `z3::ast::regexp`, caused by the only use of binop (which uses IntoAst) being behind `#[cfg(feature = "z3_4_8_14")]`.

Here I put the macro import behind the same feature flag, and make the path to `IntoAst` in the macro definition absolute to avoid it having to be present in the expanding context. I think making the path absolute is the best way to fix the conditional dependency but let me know if there's a better way. This fixes the clippy warnings for me.

### DIFF
--- a/z3/src/ast/regexp.rs
+++ b/z3/src/ast/regexp.rs
@@ -1,6 +1,5 @@
 use crate::Context;
-use crate::ast::{Ast, binop, varop};
-use crate::ast::{IntoAst, unop};
+use crate::ast::Ast;
 use std::ffi::CString;
 use z3_sys::*;
 
@@ -131,7 +130,7 @@ impl Regexp {
         }
     }
 
-    unop! {
+    crate::ast::unop! {
        /// Creates a regular expression that recognizes this regular expression one or more times (e.g. `a+`)
        plus(Z3_mk_re_plus, Self);
        /// Creates a regular expression that recognizes this regular expression any number of times
@@ -144,12 +143,12 @@ impl Regexp {
        option(Z3_mk_re_option, Self);
     }
     #[cfg(feature = "z3_4_8_14")]
-    binop! {
+    crate::ast::binop! {
         /// Creates a difference regular expression
         /// Requires Z3 4.8.14 or later.
         diff(Z3_mk_re_diff, Self);
     }
-    varop! {
+    crate::ast::varop! {
        /// Concatenates regular expressions
         concat(Z3_mk_re_concat, Self);
        /// Creates a regular expression that recognizes sequences that any of the regular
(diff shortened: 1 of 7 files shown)


lixitrixi:
The latter commit just uses an absolute path to the `binop` macro instead of conditionally importing it, as I find it a bit cleaner. I can squash them or remove the latter, whatever's preferred.

z3/src/ast/regexp.rs
```diff
@@ -144,7 +143,7 @@ impl Regexp {
        option(Z3_mk_re_option, Self);
     }
     #[cfg(feature = "z3_4_8_14")]
-    binop! {
+    crate::ast::binop! {
```
toolCHAINZ: If we're changing binop, we might also want to change unop

toolCHAINZ:
Thanks! I left one minor note; if we change the import of `binop` in that file, we should probably make `unop` use an absolute import too to be consistent.

toolCHAINZ:
I'm also curious what clippy configuration you're using; I've not seen any warnings. I would ideally like this kind of thing to be caught by CI.

I'm running `cargo clippy --all-targets` on rust `1.91.1` and it's not warning about this.

Maybe we eventually turn on clippy pedantic.

edit: ohhh wait I saw your note about features, that explains it.

lixitrixi:
I couldn't find any paths to make absolute in `unop`, but did change `trinop`.

FYI this fails when we run `RUSTFLAGS="-D warnings" cargo check --workspace --all-targets`, so not actually clippy although we do run it with similarly strict flags the step after in our CI.

Edit: I get what you mean for changing binop, sorry! I thought you meant the use of `IntoAst` without an absolute path in `trinop`.

lixitrixi:
I'm bumping into this because I'm trying to change to a submodule dependency rather than a git dependency (and since it's in the dep graph Cargo decides to give warnings about it). We'd like it to be bundled rather than use a system library. Is there a better way to do this with minimal setup?

toolCHAINZ:
Is there a reason you can't use it as a crates.io library? The latest version supports bundled building even when pulling from crates.io

toolCHAINZ:
Also, you might try the `gh-release` feature, which works a similar way but downloads the pre-built libraries from z3's github releases.

Z3 is a large piece of software that is slow to compile, so unless you very specifically want to build it from source, `gh-release` is likely to be what you want.

Looking on my own computer, it looks like the `gh-release` feature statically links `z3` too (which is actually a suprise to me), so it is really very similar, just no massive C++ build step.

PR #472, opened 2025-11-20
feat: Use native-tls-vendored feature of reqwest, removing build dependency on libssl-dev

grebnetiew:
Building `z3-sys` with the `bundled` or `gh-release` feature requires `openssl-sys` (via `reqwest`), which in turn introduces a dependency on the system package `libssl-dev` (in case of Ubuntu) or similar.

Changing the `reqwest` feature `default-tls` (which is equal to `native-tls`) to `native-tls-vendored`, makes `reqwest` use the `vendored` version of `openssl`, allowing `z3-sys` to be built on systems that do not have `libssl-dev`. This makes the package more versatile.


toolCHAINZ:
Thank you!

PR #475, opened 2025-11-24
feat: Add Sort::type_variable

toolCHAINZ:
This adds a low-level and high-level binding to `Z3_mk_type_variable`. Gated behind of a feature because of ubuntu.


toolCHAINZ:
This type is useless on its own and should probably just be added as part of a larger feature as in #540

PR #500, opened 2026-02-23
feat: algebraic numbers, polynomials, enhanced floats, AST vectors, and quantifier elimination

toolCHAINZ:
API extensions contributed by @NikolajBjorner using copilot.

### DIFF
--- a/z3-sys/src/lib.rs
+++ b/z3-sys/src/lib.rs
@@ -6420,7 +6420,12 @@ unsafe extern "C" {
     /// |  (=> atoms horn_rule)
     /// |  atom
     /// ```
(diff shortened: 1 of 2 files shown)


toolCHAINZ:
As a reminder to myself, this PR makes actual high-level bindings around Z3_ast_vector. Previously we did not have these and there's lots of finnicky low-level construction of ast_vectors in the high-level bindings. An actual high-level binding for that type itself would probably allow for simplifying the implementation of many different APIs.

toolCHAINZ:
I am now fairly happy with these changes. The ubuntu CI failure I'm guessing is a very old upstream Z3 bug as it doesn't manifest in any other run and that one is using Z3 4.8.

PR #511, opened 2026-03-04
feat: add `with` method to Tactic

longlinh123456:
Adds a binding to `Z3_tactic_using_params` so parameter sets can be provided to tactics directly.


toolCHAINZ:
This looks great, thanks!

toolCHAINZ:
This has been released as `v0.19.12`

PR #513, opened 2026-03-06
fix(z3-sys): raise GitHub download timeout for gh-release

milevin:
CI builds can fail due to intermittent network slowdowns; a 30s timeout is too short. We use Z3 in production builds and cannot assume it is present on every builder, so we rely on the `gh-release` path and need it to be resilient to transient slowness.

### DIFF
--- a/z3-sys/build.rs
+++ b/z3-sys/build.rs
@@ -89,6 +89,7 @@ fn link_against_cxx_stdlib() {
 #[cfg(any(feature = "gh-release", feature = "bundled"))]
 mod gh_release {
     use std::path::Path;
+    use std::time::Duration;
 
     use super::*;
     use reqwest::blocking::{Client, ClientBuilder};
@@ -226,7 +227,9 @@ mod gh_release {
     }
 
     pub fn get_github_client() -> Client {
-        let client = ClientBuilder::new().user_agent("z3-sys");
+        let client = ClientBuilder::new()
+            .user_agent("z3-sys")
+            .timeout(Duration::from_secs(300));
         let mut headers = HeaderMap::new();
         if let Ok(val) = env::var("READ_ONLY_GITHUB_TOKEN") {
             headers.insert(AUTHORIZATION, format!("Bearer {val}").parse().unwrap());


toolCHAINZ:
Thanks!

I'll go ahead and merge this though I'm curious about your particular setup if you don't mind answering. I'm currently mulling over making some changes to the build options this library provides (see #510) so it's helpful to know how people are using it (though don't worry, I won't be removing `gh-release`).

* Is this github actions or some other CI?
* Do you specifically rely on static-link builds of z3 (for distribution) or is the usage of gh-release here purely for ci build convenience?
* Do you use the default version of z3 that gh-release points to (right now it is pointing at 4.16.0) or are you overriding it to target an old release for any reason?

toolCHAINZ:
This has been released in `0.19.13`

milevin:
Thank you @toolCHAINZ!

Some context: I am with dbt labs and we use Z3 for generating test data given a SQL query. This uses symbolic/concrete execution and Z3 for constraint solving. (The usual spiel)

> * Is this github actions or some other CI?

It's in the build:

```
[target.'cfg(target_os = "macos")'.dependencies]
z3 = { version = "0.13.3", features = ["gh-release"] }
```

Whenever we build Fusion on a Mac (both on dev machines and in CI), we hit this potential problem

> * Do you specifically rely on static-link builds of z3 (for distribution) or is the usage of gh-release here purely for ci build convenience?

To be clear: this is linked only into an internal binary used for test generation. It is not shipped to our customers. 

We use gh-release for convenience: many devs and build machines don't have Z3 pre-installed and this avoids an extra step of `brew install z3`.

> * Do you use the default version of z3 that gh-release points to (right now it is pointing at 4.16.0) or are you overriding it to target an old release for any reason?

Default is perfectly fine.

toolCHAINZ:
Ah, thanks! This seems like a pretty reasonable-use for gh-release! The new version of the `bundled` feature, not yet released, will pull down z3 from a cached copy on crates.io and build it; definitely a different tradeoff there, removing the github requests and trading it for a higher cold-build time. If gh-release has been working for you (modulo timeouts), then that sounds great.

Do note when you update, the library has gone through a lot of changes since `0.13.3`. This is mainly removing lifetime parameters and `ctx` args from all the APIs. It should be a pretty painless migration but you will likely need code changes when you upgrade.

PR #515, opened 2026-03-12
Update zip dependency to latest

ThomasTNO:


ThomasTNO:
@toolCHAINZ this seems to work just fine locally. Can you trigger the skipped pipelines to verify?

toolCHAINZ:
I should probably configure these workflows to allow manual triggering; right now the best way will be to just merge this in. Since it worked for you locally and we're using very high-level APIs I'm just going to merge and we'll see if it passes the full CI run on `master`.

Thanks!

I'm also looking at making a change to bundled building in the near future (see #510) which will eliminate the build dep on zip.

PR #557, opened 2026-04-12
fix: align MSVC runtime library to crt-static target feature on Windows

kkent030315:
On Windows MSVC, derive `CMAKE_MSVC_RUNTIME_LIBRARY` from the `+crt-static` target feature instead of hardcoded value.


toolCHAINZ:
Looks great, thank you!

toolCHAINZ:
Released as `v416.0.2`

PR #565, opened 2026-06-22
feat: Add bindings for `get_lower`/`get_upper` in the `Optimize` solver.

daemontus:
A relatively minor change, allows extracting current optimization bounds from the `Optimize` solver using existing `z3-sys` methods.


toolCHAINZ:
Looks great, thanks!

PR #567, opened 2026-06-22
Add model handler to `Optimize`

daemontus:
In `Optimize`, it is possible to register a "model handler" that's invoked every time the solver finds a better (not necessarily optimal) solution. This PR adds the option to register such handler from safe bindings. The solution is somewhat technical because it needs to safely "forward" the C-style `fn` handler call to a rust `Fn` callback, plus the handling of `Model` instances is also a bit unusual. As far as I recall, this is inspired by a similar implementation in the Python Z3 bindings.

It is probably not ideal, but if there are some other options of how to implement this feature, I'd be happy to update the PR.

### DIFF
--- a/z3/src/lib.rs
+++ b/z3/src/lib.rs
@@ -81,7 +81,6 @@
 #![deny(missing_debug_implementations)]
 
 use std::cell::RefCell;
-use std::collections::BTreeSet;
 use std::ffi::CString;
 use z3_sys::*;
 pub use z3_sys::{AstKind, GoalPrec, SortKind};
@@ -199,8 +198,8 @@ pub struct Model {
 pub struct Optimize {
     ctx: Context,
     z3_opt: Z3_optimize,
-    // Stores IDs of model handlers associated with this `Optimize` solver.
-    registered_handlers: RefCell<BTreeSet<u32>>,
+    // Stores ID of a model handler associated with this `Optimize` solver.
+    registered_model_handler: RefCell<Option<u32>>,
 }
 
 /// Context for Horn clause / Datalog solving.
--- a/z3/src/lib.rs
+++ b/z3/src/lib.rs
@@ -80,6 +80,8 @@
 #![warn(clippy::doc_markdown)]
 #![deny(missing_debug_implementations)]
 
+use std::cell::RefCell;
+use std::collections::BTreeSet;
 use std::ffi::CString;
 use z3_sys::*;
 pub use z3_sys::{AstKind, GoalPrec, SortKind};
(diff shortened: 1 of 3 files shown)


toolCHAINZ:
Thanks for the contribution(s)! Just wanted to flag that I'd seen this but haven't had time to look at it in detail, but I hopefully will soon.

There's a few Z3 features requiring callbacks missing from the high level bindings (e.g. the user propagator), so as I look through this I'm wanting to think through whether the strategy used here would be appropriate for these other cases as well, as I'd like to avoid duplication in this kind of finicky FFI code.

Hopefully I'll get a chance to glance at this soon.

daemontus:
Sure, no problem, take your time :)

Honestly, I was a bit unsure regarding how to approach this. Having the option to only register `fn` handlers seemed quite limited, plus there was the need to keep track of the `Model` somehow. But this solution also seems somewhat over-engineered to me. I don't have that much experience interfacing Rust with C APIs, so there might be a better way that I am missing.

toolCHAINZ:
I finally gave this a look. Agreed, I think the thread_local stuff was probably overbuilt: none of the Z3 types are Send/Sync anyway so IDK if that actually buys anything.

Attempted an implementation instead using a shared generic struct (to try to establish a reusable pattern for different callback-based FFI endpoints), and boxed closures (w per-callback API trampolines).

I don't know if I'm totally happy with this just yet (this is the result of a several-shot impl + refinement with claude code). I think the actual FFI piece is fine but I think there might be more to do to improve the "shared abstraction" story.

I also make the simplifying design choice to say that model handlers are not preserved through `Translate` calls; this seemed like a tricky soundness area that I didn't want to think about for an initial implementation, so right now they are explicitly not carried over.

I wasn't able to push to your branch, so I pushed to https://github.com/prove-rs/z3.rs/compare/master...add-optimize-model-handler if you'd like to look it over.

I made https://github.com/prove-rs/z3.rs/pull/577 with my changes.

toolCHAINZ:
To prove out the design in my extensions to your work, I used claude code to quickly prototype a user propagator implementation using this same callback abstraction (and implementing the general pattern I suggested in https://github.com/prove-rs/z3.rs/pull/344#issuecomment-3236279448).

It seems to work nicely and the code looks pretty clean to me.  I did not have to alter the callback FFI design at all to accommodate it, so I think I may be OK merging it (and maybe then separately merging the new user propagator bindings).

daemontus:
Thank you! The new solution seems indeed much cleaner. As such, I am closing this in favor of #577.

PR #568, opened 2026-06-26
fix(Real): from_rational must not truncate i64 args to i32

progwriter:
Real::from_rational takes i64 num/den but lowered them via Z3_mk_real, whose C signature is (int, int). Casting the i64 args to c_int silently truncated any value outside the i32 range: e.g. 634909090909091/100000000000 (= 6349.09090909091) wrapped to ~1.0326, producing a wrong-but-valid Real with no error. This corrupts any consumer building rationals with large numerators/denominators (common when a decimal like 6349.09090909091 is expressed as digits/10^k).

Use Z3_mk_real_int64(num, den), the int64 fraction API, which preserves the full i64 range. Negatives still work (covered by algebraic_tests' from_rational(-7,4)).

Adds a regression test asserting a >i32 rational lowers exactly and not to the old truncated value.

### DIFF
--- a/z3/src/ast/real.rs
+++ b/z3/src/ast/real.rs
@@ -66,12 +66,10 @@ impl Real {
         let ctx = &Context::thread_local();
         unsafe {
             Self::wrap(ctx, {
-                Z3_mk_real(
-                    ctx.z3_ctx.0,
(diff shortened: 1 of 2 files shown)


toolCHAINZ:
Great catch, thank you!

PR #572, opened 2026-07-22
chore!: updated z3-sys generated bindings for z3 5

toolCHAINZ:


toolCHAINZ:
This is probably going to break CI, and CI will stay broken until Z3 5.0.0 gets added to vcpkg and brew.

PR #577, opened 2026-07-27
feat: Optimize::set_model_handler

toolCHAINZ:
Implements part of https://github.com/prove-rs/z3.rs/issues/544

### DIFF
--- a/z3/src/optimize.rs
+++ b/z3/src/optimize.rs
@@ -1,10 +1,12 @@
 use log::debug;
 use std::borrow::Borrow;
+use std::cell::RefCell;
+use std::collections::BTreeMap;
 use std::convert::TryInto;
-use std::ffi::{CStr, CString};
+use std::ffi::{CStr, CString, c_void};
 use std::fmt;
 use std::iter::FusedIterator;
-
+use std::sync::atomic::AtomicU32;
 use z3_sys::*;
 
 #[cfg(feature = "z3_4_16")]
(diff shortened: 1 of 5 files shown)


z3/src/optimize.rs
```diff
@@ -379,6 +410,94 @@ impl Optimize {
...
+            Z3_optimize_register_model_eh(
+                self.ctx.z3_ctx.0,
+                self.z3_opt,
+                z3_model,
+                new_nn.as_ptr().cast::<c_void>(),
+                Some(model_handler_trampoline),
+            );
+
+            // Swap in the new pointer and free any previous handler state.
+            if let Some(old_nn) = self.handler.replace(Some(new_nn)) {
```
daemontus: This is vey much a nitpick, but maybe this and `clear_model_handler` are a bit redundant... Something like `replace_model_handler(&self, Option<NonNull<ModelHandlerState>>)` that safely drops the *previous* model handler could be called from `drop` and from `set_model_handler`. But it's a small change... my only motivation is minimizing the number of unsafe code lines.

daemontus:
I looked at the whole implementation. It looks very nice to me. I left one minor nitpick, but otherwise I haven't found anything that would seem questionable to me :) Thank you for preparing this!

z3/src/optimize.rs
toolCHAINZ: Hmmmm, that's a good point, that might be a cleaner way to do this. I'll have a look when I get the chance and push up an update. Thanks for the review!

toolCHAINZ: Implemented in [a24ef8c](https://github.com/prove-rs/z3.rs/pull/577/commits/a24ef8c0bc88e653c93fe8be21976de873f9db21)

PR #583, opened 2026-08-27
feat!: auto-detect z3 version

toolCHAINZ:
Looking over a bunch of the to-be-added features it struck me that the versioning strategy in this library was about to get a lot more complicated.

Previous to 0.20.0, the z3 crate basically just implemented the API as it existed at 4.8.17, with no existing utilities to allow adding support for newer APIs.

The changes in 0.20.0 allowed better explicit tying of z3-rs versions to z3 versions, but committed bindgen results, so it was possible for bindgen to get out-of-sync vs the linked library on enum definitions. Additionally, the current solution for properly gating new apis (a cargo feature per-z3 version) was fine for the _one_ feature that used it, but if we add all these other missing APIs, we suddenly have many different versions to track and this gets unwieldy.

This PR is a claude-coded (and will probably be hand-tweaked once I have time to thoroughly review it) implementation of the following strategy:

* z3-sys, since it always has either `z3-version.h` available or has a pre-set z3 version given to it (via e.g. `gh-release`), emits cargo cfgs identifying the version number
* z3-sys gains metadata defining which version its generated enums were built against, prompting the user with a warning to enable the bindgen feature if they are using a different version of z3.
* z3 gains a build.rs that further sets "z3 is greater or equal to version x `cfg`s" based on the reported version number from z3-sys. These `cfg`s, which are maintained in one place and auto-set, are then used to gate APIs that need them.
* The library's minimum version is now assumed to be z3 4.13.3, the version shipping with ubuntu 26.04. Older versions will give a build.rs-sourced error (more digestible than a linker error). Newer features will be supported behind cfg guards. The library will assume the standard user is using z3 5.0 (so 4.13.3 will work but will give an error due to enum mismatch between that version and 5 unless bindgen is re-run).

I have yet to play with this, and need to try it across different platforms and make sure it is actually doing what I wanted it to, but I think this would be a much cleaner version-selection story than what's there now.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -51,7 +51,6 @@ jobs:
         run: cargo test --workspace --features num
 
   build_on_wasm:
-    if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v') || github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
@@ -92,7 +91,6 @@ jobs:
           CMAKE_CXX_COMPILER_LAUNCHER: "sccache"
 
   build_with_vendored_z3:
-    if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v') || github.event_name == 'workflow_dispatch'
     strategy:
       matrix:
         build: [linux, macos, windows]
@@ -137,7 +135,6 @@ jobs:
           SCCACHE_MSVC_Z7: "1"
 
   build_with_vcpkg_installed_z3:
-    if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v') || github.event_name == 'workflow_dispatch'
     strategy:
       matrix:
         build: [linux, macos, windows]
@@ -203,7 +200,6 @@ jobs:
         run: cargo clippy --workspace --all-targets --features num
 
   build_with_gh_release_z3:
-    if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v') || github.event_name == 'workflow_dispatch'
     strategy:
       matrix:
         build: [linux, macos, windows]
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -51,6 +51,7 @@ jobs:
         run: cargo test --workspace --features num
 
   build_on_wasm:
+    if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v') || github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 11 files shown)


toolCHAINZ:
Pretty slick:

Running tests on a z3 version _without_ `Optimize::translate` support:

```
~/R/z3.rs ❯❯❯ Z3_SYS_Z3_VERSION=4.15.0 cargo test -p z3 --no-default-features --features gh-release --test optimize_translate --test optimize_model_handler
    Finished `test` profile [unoptimized + debuginfo] target(s) in 2.05s
     Running tests/optimize_model_handler.rs (target/debug/deps/optimize_model_handler-793aee1a2116763c)

running 5 tests
test handler_fires_at_least_once ... ok
test handler_fires_multiple_times_for_improving_sequence ... ok
test drop_with_active_handler_does_not_panic ... ok
test handler_receives_non_empty_model ... ok
test second_set_model_handler_replaces_first ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

     Running tests/optimize_translate.rs (target/debug/deps/optimize_translate-be732c881ec4b6ad)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

Running against 4.16.0, in which this feature is present:

```
~/R/z3.rs ❯❯❯ Z3_SYS_Z3_VERSION=4.16.0 cargo test -p z3 --no-default-features --features gh-release --test optimize_translate --test optimize_model_handler
    Finished `test` profile [unoptimized + debuginfo] target(s) in 28.80s
     Running tests/optimize_model_handler.rs (target/debug/deps/optimize_model_handler-793aee1a2116763c)

running 10 tests
test cloned_check_with_no_handler_after_original_drops ... ok
test handler_fires_at_least_once ... ok
test handler_fires_multiple_times_for_improving_sequence ... ok
test handler_receives_non_empty_model ... ok
test second_set_model_handler_replaces_first ... ok
test drop_with_active_handler_does_not_panic ... ok
test drop_translated_instance_with_no_handler_does_not_corrupt_original ... ok
test translate_does_not_carry_handler_across_contexts ... ok
test translated_check_with_no_handler_after_original_drops ... ok
test translate_with_no_handler_gives_no_handler ... ok

test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s

     Running tests/optimize_translate.rs (target/debug/deps/optimize_translate-be732c881ec4b6ad)

running 2 tests
test test_optimize_clone ... ok
test test_optimize_translate ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s
```

So the auto-gating seems to be working.

toolCHAINZ:
Looks like pkg-config on linux/macos isn't working yet though: https://github.com/prove-rs/z3.rs/actions/runs/33063909852/job/98489184754?pr=583

These runs don't see any gated tests, indicating that the cfg isn't set, further indicating the version parsing must have failed and it fell back to 4.13.3 support.

edit: I think this is effectively fixed in [c41e95f](https://github.com/prove-rs/z3.rs/pull/583/commits/c41e95f6ec5e78ddfe485d8f1ae052c6dae21729)

PR #586, opened 2026-08-29
feat: add FiniteSet sort and operations (Z3 5.0)

### LINKED ISSUE #576: feat: Add FiniteSet sort and operations (Z3 5.0)
## Summary

Z3 5.0 introduced a `FiniteSet` sort distinct from the existing array-based `Set`.
Finite sets have dedicated operators including cardinality, higher-order map/filter,
and integer range construction — none of which are available on the existing `Set`.

## Z3 APIs

**Sort**
- `Z3_mk_finite_set_sort(c, elem_sort)` — create a finite set sort
- `Z3_is_finite_set_sort(c, s)` — check if a sort is a finite set sort
- `Z3_get_finite_set_sort_basis(c, s)` — get the element sort

**Constructors**
- `Z3_mk_finite_set_empty(c, set_sort)` — empty set
- `Z3_mk_finite_set_singleton(c, elem)` — singleton set
- `Z3_mk_finite_set_range(c, low, high)` — integer range [low, high]

**Set operations**
- `Z3_mk_finite_set_union(c, s1, s2)` — union
- `Z3_mk_finite_set_intersect(c, s1, s2)` — intersection
- `Z3_mk_finite_set_difference(c, s1, s2)` — difference

**Predicates**
- `Z3_mk_finite_set_member(c, elem, set)` — membership (returns Bool-valued AST)
- `Z3_mk_finite_set_subset(c, s1, s2)` — subset check (returns Bool-valued AST)

**Higher-order**
- `Z3_mk_finite_set_size(c, set)` — cardinality (returns Int-valued AST)
- `Z3_mk_finite_set_map(c, f, set)` — apply function to all elements
- `Z3_mk_finite_set_filter(c, f, set)` — filter by predicate

## Notes

- `FiniteSet` cannot be built up element-by-element (no `add`/`del`); sets are
  constructed via `singleton`, `union`, `range`, etc.
- Natural home: a new `ast::FiniteSet` type in `z3/src/ast/finite_set.rs`, mirroring
  the structure of `ast::Set`
- `DeclKind` enum already gains 13 new variants in `z3-sys` 0.12.0 (all `FiniteSet*`)
- `Z3_mk_finite_set_map` takes a lambda/function AST — may need a note about how to
  construct one (e.g. via quantifiers or `rec_func_decl`)

toolCHAINZ:
## Summary
- Adds `ast::FiniteSet`, mirroring `ast::Set`'s structure: `new_const`/`fresh_const`, `empty`, `singleton`, `range`, `member`, `union`, `intersect`, `difference`, `subset`, `size`, `map`, `filter`.
- Adds `Sort::finite_set`, `Sort::is_finite_set`, `Sort::finite_set_basis`, and `Dynamic::as_finite_set`.
- `FiniteSet` doesn't support `add`/`del` (finite sets aren't built element-by-element in Z3's API).

## Design decisions
- All required z3-sys FFI bindings already existed (`z3-sys/src/generated/functions.rs`), so this is purely `z3` crate work.
- Gated the whole feature behind `#[cfg(z3_5_0_0)]`, reusing the version-threshold mechanism already established for `Optimize::solutions()` (`z3/build.rs`), since `FiniteSet` was only added in Z3 5.0.
- `map`/`filter` take a typed `&Array` rather than a generic `Ast`/`Dynamic`. Z3 represents function values structurally as arrays (confirmed via `z3-src/z3/src/test/finite_set.cpp`), and this crate already produces function-valued terms as `Array` elsewhere (`ast::Set`, `ast::lambda_const`), so this catches sort mismatches at compile time instead of deferring to a Z3 runtime error.
- The `map`/`filter` test only checks construction/sort correctness, not satisfiability: as of Z3 5.1.0, asking the solver to reason about the *contents* of a `set.map`/`set.filter` result hangs indefinitely (verified independently against the `z3` CLI via raw SMT-LIB2, with `timeout` param set), so this isn't a bindings bug — just a current limitation of Z3's FiniteSet theory.

## Test plan
- [x] `cargo test -p z3` (unit + integration + doctests) — all passing
- [x] `cargo clippy -p z3 --all-targets -- -D warnings` — clean
- [x] `cargo fmt -p z3 -- --check` — clean

Closes #576


toolCHAINZ:
I need to look over the map/filter logic here. I have not used the new FiniteSet API yet so I'm not sure if this implementation makes sense.

PR #592, opened 2026-09-12
Improve windows cross compilation support and build non-msvc windows …

MI3Guy:
…targets.

Full disclosure: This change was mostly made and tested using codex. I compiled and tested this using `cargo zigbuild` on Linux for `x86_64-pc-windows-gnu` and `x86_64-pc-windows-gnullvm`. I don't have a msvc toolchain installed so I have not tested against that (does CI cover that?).

This change should enable cross compilation and compilation for non-MSVC windows targets.


toolCHAINZ:
Thanks for the contribution and disclosure! No issue on my end using codex (I too have taken to using coding agents extensively) provided that the changes are human-motivated and understood.

Yes, I believe the windows CI job is using MSVC, so this should be exercised by the windows vendored job. I do have those jobs disabled on pull requests though since when there's cache misses they take a very long time to complete. I can look into allowing them to be selectively run for branches like this.

Out of curiosity, do non-windows cross builds (e.g. Linux arm from Linux x64) work with zigbuild? I've not used it myself yet, and to my knowledge no previous work on this library has really targeted cross builds as an explicit feature, so I'd be interested to hear if it "just works" for any targets.

I definitely do see value in enabling this if it doesn't break existing windows support. I'll look into enabling that job, read a little about zigbuild and then look this over when I get a chance.

Thanks again!

MI3Guy:
Yes, I've used zigbuild to cross compile for a different CPU on Linux as well for a project that uses this crate. The only issues I ran into were when building for Windows. One nice feature of using zigbuild is you can pin the glibc version on Linux builds without needing to have a sysroot set up for that version.

z3-src/src/lib.rs
```diff
@@ -102,23 +104,29 @@ fn build_cmake(src_dir: &Path) -> PathBuf {
...
-    if cfg!(target_os = "windows") {
-        // -MP enables parallel MSVC compilation, but conflicts with compiler launchers
-        // like sccache which handle parallelism per-invocation and cannot cache -MP mode.
-        if env::var("CMAKE_CXX_COMPILER_LAUNCHER").is_err() {
-            cfg.cxxflag("-MP");
-            cfg.build_arg("-m");
-        }
+    if env::var("CARGO_CFG_TARGET_OS").as_deref() == Ok("windows") {
         cfg.cxxflag("-DWIN32");
         cfg.cxxflag("-D_WINDOWS");
```
toolCHAINZ, line 109: I'm curious as to the purpose of these lines. Are these not already defined by default?

toolCHAINZ (commented):
These changes look reasonable to me, with one small question.

It looks like the MSVC windows CI also still passes with this change: https://github.com/prove-rs/z3.rs/actions/runs/35273427264

z3-sys/build.rs
```diff
@@ -235,7 +235,7 @@ mod gh_release {
             println!("Found cached z3 at {}", z3_dir.display());
         }
 
-        let lib = if cfg!(target_os = "windows") {
+        let lib = if target_os == "windows" {
```
toolCHAINZ, line 242: I'm trying to think through whether there is a combination of build options that will lead down this path while using mingws linker

Z3 only distributed a MSVC build. Is there a way a confused user might attempt to use zigbuild and `gh-release` and then get a weird linker error when it can't link `libz3.lib`? If so that would ideally be a cargo warning or a panic or something. But I also recognize this is pretty niche and people are unlikely to hit it.

z3-src/src/lib.rs
MI3Guy: Worth noting, those lines weren't a change I made. They were already present.

That said, I wouldn't be surprised if the were unnecessary.

z3-sys/build.rs
MI3Guy: I think you are right about this. If we try to compile for windows-gnu, this will cause an error. I'll see if I can replicate this.

MI3Guy:
I also noticed that z3-syz depends on the published version of z3-src rather than using a path dependency. I assumed this was intentional, but it also means that CI won't be building against the in-repo version of z3-src. Do you want me to change this as well?

z3-sys/build.rs
MI3Guy: I learned something interesting. Apparently, since it is only using the C ABI, it is actually able to link against the .lib and load the gh-release DLL. This means that you could actually support this scenario. That means that we have odd conditionals to use the msvc naming when building, but since it does appear to actually work,  

```
#[cfg(all(not(feature = "gh-release"), target_env = "msvc"))]
#[link(name = "libz3")]
```

I was able to run the unit test executable (using wine) and it works. I've pushed a change that makes the linking work. That said, this is a bit of an odd scenario, so I could also roll this back and have it produce an error if you think that's better.

z3-src/src/lib.rs
toolCHAINZ: Ah I misread the diff! Disregard. I can look separately later whether these actually matter.

z3-sys/build.rs
toolCHAINZ: Oh interesting! With the changes you pushed in, it looks like we would never emit the link directive under gh-release (since it's guarded behind "not" for both instances).

Since this is niche, maybe we should roll this back in the interest of simplicity (and explore it in a follow-up if there's a reason to support it).

toolCHAINZ:
> I also noticed that z3-syz depends on the published version of z3-src rather than using a path dependency. I assumed this was intentional, but it also means that CI won't be building against the in-repo version of z3-src. Do you want me to change this as well?

Ah yeah, this was this way because having the path dependency led to test failures when bumping the vendored version of Z3 when there was an upstream enum redefinition requiring a bindgen rerun. Since I had only just made z3-src, I was wanting to publish it separately from the other crates and needed passing CI.

I don't think that separation is necessary anymore so I'm open to re-adding the path dependency.

I should note, one other thing this gives us: z3 src does not bundle the whole source tree; it omits z3 examples and several other unneeded source files (e.g. other language APIs). Building against the published version ensures that we are bundling all the necessary files in the src crate. This was useful when I added the crate initially, but this is unlikely to change and works as it is. So idk if this is really a valuable check anymore

z3-sys/build.rs
toolCHAINZ: Ohhh I just saw the verbatim link. I'll need to read more about this. Still tempted to roll back for simplicity

MI3Guy: Yeah, I wasn't sure which way to go. Let me know and I'll remove that change.

toolCHAINZ: Hmm yeah let's roll this back for now; can re-add it if it comes up.

I think we should also do the crate path dep that you mentioned if you don't mind also applying that.

MI3Guy: Updated, squashed onto the latest to simplify commit history.

toolCHAINZ (approved):
Looks great, thanks! This is a cool addition!

PR #578, opened 2026-08-12
Fix UAF bug in ApplyResult

Ollie-Pearce:
The `ApplyResult` type contains a raw pointer to a C++ memory object in libz3 and has a derived `Clone` implementation. If `Clone` is called on `ApplyResult` a reference to the libz3 memory object is created without incrementing the associated reference counter. 

```rust
#[derive(Clone, Debug)]
pub struct ApplyResult {
    ctx: Context,
    z3_apply_result: Z3_apply_result,   // Pointer to libz3 object
}
```

When `Drop` is called on an instance of `ApplyResult` the reference counter is decremented. Creating an instance of `ApplyResult`, cloning that instance and then dropping both the clone and original causes a UAF. This bug can be triggered from safe code. 

```rust
impl Drop for ApplyResult {
    fn drop(&mut self) {
        unsafe {
            Z3_apply_result_dec_ref(self.ctx.z3_ctx.0, self.z3_apply_result);
        }
    }
}
```

### Test case
```rust
use z3::{Goal, Tactic};

#[test]
fn clone_then_drop_both_owners() {
    let goal = Goal::new(false, false, false);
    let a = Tactic::new("skip").apply(&goal, None).unwrap();
    let b = a.clone();
    drop(a);
    drop(b);
}
```

### Verification
```
cargo test -p z3 --test apply_result_clone_drop --no-run
valgrind --leak-check=no ./target/debug/deps/apply_result_clone_drop-21a20d95342baf41
```

### Valgrind output:
```
==65755== Thread 2 clone_then_drop:
==65755== Invalid read of size 4
==65755==    at 0x547012E: ??? (in /usr/lib/x86_64-linux-gnu/libz3.so.4)
==65755==    by 0x547044E: Z3_del_context (in /usr/lib/x86_64-linux-gnu/libz3.so.4)
==65755==    by 0x18444D: <z3::context::ContextInternal as core::ops::drop::Drop>::drop (context.rs:17)
==65755==    by 0x183F0A: core::ptr::drop_in_place<z3::context::ContextInternal> (mod.rs:805)
==65755==    by 0x1851A1: alloc::rc::Rc<T,A>::drop_slow (rc.rs:396)
==65755==    by 0x18428B: <alloc::rc::Rc<T,A> as core::ops::drop::Drop>::drop (rc.rs:2472)
==65755==    by 0x183FFA: core::ptr::drop_in_place<alloc::rc::Rc<z3::context::ContextInternal>> (mod.rs:805)
[...]

PR #594, opened 2026-09-20
fix: UAF bug in Goal

Ollie-Pearce:
The `Goal` type contains a raw pointer to an external memory object which can be duplicated via `Clone` without incrementing the reference counter. 

If an instance of `Goal` is created, duplicated via `Clone` and then both instances are dropped a UAF occurs.

You should be able to reproduce the bug with this test case:

```rust
fn clone_then_drop_both_owners() {
    let a = Goal::new(false, false, false); 

    let b = a.clone();
    drop(a);
    drop(b);
}
```

## Fix:

Use `wrap` for `Goal`'s `Clone` implementation. This is hinted at in the "todo" comment and you can see the same pattern in [this similar PR I made](https://github.com/prove-rs/z3.rs/pull/578)


toolCHAINZ:
Thanks! I looked for other instances of this pattern last time but somehow missed this.

